PR #5, opened 2015-01-07
patch for multple expressions

randy3k:
If an expression contains multiple expressions, we should not only only evaluate the first expression.

```
julia> using RCall

julia> Reval(Rparse("a=1;b=2"))
RCall.SEXP{14}(Ptr{Void} @0x00007fbc8356fcf8)

julia> R.str(Reval(:a))
 num 1
RCall.SEXP{0}(Ptr{Void} @0x00007fbc86813b78)

julia> R.str(Reval(:b))
Error: object 'b' not found
```


dmbates:
Thanks for the pull request.  I was planning to make the list (VECSXP) and character string (STRSXP) types behave as iterators and will do the same for the expression (EXPRSXP) type.

randy3k:
It could be a little bit tricky for `VECSXP` since there are several types of getters `x[1],  `x[[1]]` and `x$foo`.

PR #18, opened 2015-01-30
bugfix for sexp(::Matrix) the num of elements to copy was not correctly ...

robertfeldt:
...calculated

Only fixed for 2-dim case though. Seems to be wrong also for higher dim cases so should be fixed later...

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -76,17 +76,17 @@ lang3(s1::SEXP,s2::SEXP,s3::SEXP) =
 
 @doc "Create a 3-argument function call from a symbol and the arguments"->
 lang4(s1::SEXP,s2::SEXP,s3::SEXP,s4::SEXP) =
-    sexp(ccall((:Rf_lang3,libR),Ptr{Void},
+    sexp(ccall((:Rf_lang4,libR),Ptr{Void},
                  (Ptr{Void},Ptr{Void},Ptr{Void},Ptr{Void}),s1,s2,s3,s4))
 
 @doc "Create a 4-argument function call from a symbol and the arguments"->
 lang5(s1::SEXP,s2::SEXP,s3::SEXP,s4::SEXP,s5::SEXP) =
-    sexp(ccall((:Rf_lang3,libR),Ptr{Void},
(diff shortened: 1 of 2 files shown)


robertfeldt:
PR now also includes the fix for lang4, lang5 and lang6. They were calling lang3.

dmbates:
Thanks for the pull request - I somehow always manage to get dumb cut-and-paste errors in my code.  Sorry for the delay in responding to the PR.  I was on a road trip for the last week or so.

robertfeldt:
No worries. Thx for RCall by the way. I find it easier to get going than with Rif.jl. OTOH I get some strange errors with repeated calls to the kernlab package. It's weird because it happens after some time and quite irregularly as to when. Some mem leak or something. I'll get back to you if I can pinpoint it with a clean example. Not sure about that though... ;)

dmbates:
I imagine the problem with the `kernlab` package is related to calling `R_PreserveObject` too frequently in the interface.  It would be best to have an option for copying the contents of an R object into storage allocated by Julia instead of preserving the storage allocated by R.  Calling `R_PreserveObject/R_ReleaseObject` is fine for a few values but not for large numbers of values because of the way that these functions in the R interface are implemented.

Of course it may be something else entirely.  Memory management within R is very tricky.  Part of my motivation in switching to Julia was to avoid the numerous pitfalls of writing compiled code to interface with R.

robertfeldt:
Ah, that sounds plausible, thx for the hint. Anything I can do to avoid it? I have a number of calls to vec in code that I loop over and I see vec calls R_PreserveObject. Hmm...

dmbates:
I think the solution is to create a version of `vec` that copies the R array into a Julia array.  That way Julia can do the garbage collection and the R object does not need to be preserved.  I'll create an issue about this to get more comments.

robertfeldt:
Do you mean something like this:

https://github.com/robertfeldt/RCall.jl/commit/a092dd1cddc291d92583dd14d9626d7e5b400b6a

robertfeldt:
I didn't dare use a copy or copy! so went with a "manual" copy. Will try and see if this can help alleviate my kernlab problems...

PR #20, opened 2015-02-02
Update README.md

cmey:
fixed typo in Julia code : Pkg.add("RCall.jl") doesnt work, but Pkg.add("RCall") works


dmbates:
Thanks.

PR #28, opened 2015-03-10
allow use of symbols in `lang` and add tests for rcall and @rimport

randy3k:
check the changes, they are self-explained.


dmbates:
Thanks again.

PR #29, opened 2015-03-11
check bounds when getting object from an environment

randy3k:
as title

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -46,13 +46,14 @@ module RCall
   Start an embedded R and create global values from several built-ins.
   In particular, globalEnv must be defined if any R expression is to be evaluated.
 """->
-function __init__()                     
+function __init__()
     argv = ["Rembed","--silent","--no-save"]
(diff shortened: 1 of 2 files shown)


randy3k:
`R_UnboundValue` is returned if one tries to get a undefined object in the environment.

```
julia> using RCall

julia> x = globalEnv[:x]
SymSxp(16777217,Ptr{Void} @0x00007fa43436dd78,Ptr{Void} @0x00007fa43436dd40,Ptr{Void} @0x00007fa43436dcd0,Ptr{Void} @0x00007fa43436dd78,Ptr{Void} @0x00007fa43436dd40,Ptr{Void} @0x00007fa43436dd78)
```

The object `R_UnboundValue` was wrapped by `sexp` to become a `SymSxp`.
It is not a promise, and in fact, a segfault error will be thrown if we evaluate this `SymSxp`. 
**Be caution, the following will kill your julia**
```
reval(x)
```

PR #30, opened 2015-03-11
export package contents as LangSxp and make them callable in v0.4.x

randy3k:
as title, in v0.4.x, now it is possible to call the functions directly.
```
julia> @rimport MASS as mass

julia> rcopy(mass.ginv(sexp([1 2;3 4])))
2x2 Array{Float64,2}:
 -2.0   1.0
  1.5  -0.5
```


dmbates:
Thanks for creating this and sorry for the delay in looking at the PR.

PR #31, opened 2015-03-11
some fixes for v0.4.x

randy3k:
1. `sexp(Array{T<:Real,1},)` is ambiguous with `sexp(Array{T<:Bool,N},)`
defining `sexp(Array{Bool,1},)` first would resolve the issue

2. we have to use `unsafe_convert` to convert julia object to a pointer.


dmbates:
Thanks again.

Issue #32, opened 2015-03-11
NA values are not handled probably


randy3k:
```
julia> DataArray(reval("c(NA,1)"))
2-element DataArray{Float64,1}:
 NaN  
   1.0
```
the correct result should be similar to
```
julia> @data([NA,1.0])
2-element DataArray{Float64,1}:
  NA
 1.0
```

Error is reported when converting a vector of string containing NA.
```
julia> DataArray(reval("c(NA,'a')"))
ERROR: `bytestring` has no method matching bytestring(::CharSxp)
 in isNA at /Users/Randy/.julia/v0.3/RCall/src/sexp.jl:148
 in DataArray at /Users/Randy/.julia/v0.3/RCall/src/sexp.jl:150
```

dmbates:
I think there are two issues here: recognizing the NA value for Float64 and recognizing it for CharSxp.

dmbates:
I realized what the problem with the Float64 vectors is.  I used == on Float64 values to check if it matches R_NaReal but equality comparisons don't return true for Float64 values that correspond to NaNs.  I changed the comparison to === and also fixed the isNA method for Complex128

randy3k:
With me last PR #34, there is still a bug: not handling 'NA' and NA correctly.

```
julia> DataArray(reval("c(NA,'NA','foo')"))
3-element DataArray{ASCIIString,1}:
 NA   
 NA   
 "foo"
```

randy3k:
31fa1fe6c88998c8035aba13b6c442252556a80f fixes.

PR #34, opened 2015-03-19
dirty fix for string array containing NAs

randy3k:
As `rcopy("NA")` returns a string `"NA"`, this dirty fix will work for most cases. But it doesn't differentiate between the real `NA` and the string `"NA"`.

Related to #32


dmbates:
I'm not sure I understand the rationale.  `R_NaString` is a `CharSxp` with contents "NA".  The reason I would want to use `bytestring(R_NaString)` instead of comparison to `"NA"` is in case that string ever changed.

Am I missing what the purpose of the pull request is?  In other words, am I doing the comparison wrong or just inefficiently?

randy3k:
There is no method of `bytestring` on `CharSxp`. If I understand your rationale, you wanted to compare it with `rcopy(R_NaString)`, which is `"NA"`.

dmbates:
I would prefer the comparison to be against `rcopy(R_NaString)' then.

randy3k:
alright, I will make the change +1

PR #40, opened 2015-04-07
Change sexp tag for real values to 14 (based on Rinternals.h).

### LINKED ISSUE #39: Float64 converting to complex when using globalEnv
Fantastic package! I was hoping to wrap some R statistical functionality from Julia and this is perfect. One issue I have encountered follows (in Julia):

    using RCall
    y = randn(10)
    globalEnv[:y] = y
    rcopy("typeof(y)")

returns "complex". One can easily get around this by inserting the additional line:

    reval("y <- Re(y)")

But I thought I would raise it as an issue here just in case there are situations where this work-around will fail.

Cheers.

simonbyrne:
Fixes #39, I think.

I don't really understand the library that well, so @dmbates you should probably check this is correct.

### DIFF
--- a/src/sexp.jl
+++ b/src/sexp.jl
@@ -207,7 +207,7 @@ for (typ,rtyp,rsnm) in ((:Bool, :Int32, "Logical"),
     @eval sexp(v::$typ) = preserve(sexp(ccall(($(string("Rf_Scalar",rsnm)),libR),Ptr{Void},($rtyp,),v)))
 end
 
-for (typ,tag) in ((:Bool,10),(:Complex,15),(:Integer,13),(:Real,15))
+for (typ,tag) in ((:Bool,10),(:Complex,15),(:Integer,13),(:Real,14))
     @eval begin
         function sexp{T<:$typ}(v::Vector{T})
             vv = preserve(sexp(ccall((:Rf_allocVector,libR),Ptr{Void},(Cint,Int),$tag,length(v))))
@@ -232,7 +232,7 @@ end
 
 ## To get rid of ambiguity, first define `sexp` for array with definite dimensions
 ## then arbitrary dimensions.
-for (typ,tag) in ((:Bool,10),(:Complex,15),(:Integer,13),(:Real,15))
+for (typ,tag) in ((:Bool,10),(:Complex,15),(:Integer,13),(:Real,14))
     @eval begin
         function sexp{T<:$typ}(a::Array{T})
             rdims = sexp([size(a)...])


dmbates:
It should be 14, yes.  I had this fixed and, in fact, had the code cleaned up considerably but then tried to publish a new version of another package.  As always happens to me when I use `Pkg.publish` I ended up in a git-maze from which there was no escape.  The `Pkg.publish` call fails, apparently because I use more than one computer, and after that everything turns to custard.  The only reliable way I have of getting out of such a mess is to blow away all traces of Julia and Julia packages on my computer and start over from scratch.  Unfortunately, this resulted in my losing all my changes.

I have reached the conclusion that I cannot maintain Julia packages in METADATA, especially with the current difficulties in maintaining a version that can be used with both v0.3 and v0.4, so it would be better to remove my packages, including this one.

simonbyrne:
I can sympathise, I've long given up on `Pkg.publish`, and now just manually open pull requests on METADATA after running `Pkg.tag`.

By the way, I did appreciate the "Jon Snow convention" in the notebook.

dmbates:
@simonbyrne Do you fork a copy of 'github.com/JuliaLang/METADATA.jl' when you push the changes from the tag?  And do you destroy your copy after the pull request is accepted?  I always have this fear that somehow my forked copy of `METADATA.jl`being behind `JuliaLang/METADATA.jl` will result in cruft in the pull request.  And it doesn't help that everything is taking place on the `metadata-v2` branch.

Essentially I haven't been able to come up with a foolproof method of releasing a new version of a package that doesn't involve about 20 steps.  A two-minute change in a package takes the rest of the day to get a new version of the package released.

I was wondering if anyone would pick up on the "Jon Snow convention".

simonbyrne:
I have a fork of METADATA under my own profile, and I've set up a remote called `mine`:
```
$ git remote -v
mine	git@github.com:simonbyrne/METADATA.jl.git (fetch)
mine	git@github.com:simonbyrne/METADATA.jl.git (push)
origin	git://github.com/JuliaLang/METADATA.jl (fetch)
origin	git@github.com:JuliaLang/METADATA.jl.git (push)
```

What I do, which seems to work, is:

1. `Pkg.tag` the release.
2. run `git push mine metadata-v2` in the METADATA directory.
3. manually open a pull request
4. run `git push --tags` in the package directory (I usually do this when the pull request is merged).

Of course, this only lets you update 1 package at a time, otherwise you have to manually fork METADATA (which makes things confusing).

dmbates:
Isn't there a problem with your fork of METADATA.jl getting behind `JuliaLang/METADATA.jl` by dozens of commits or is that overcome by the `git push mine metadata-v2`?

Do you run `Pkg.update()` before `Pkg.tag`?  I have this pathological fear that somehow the three different repositories will end up in an inconsistent state and everything will blow up in my face (as has happened frequently).  Someone will accept a pull request on `JuliaLang/METADATA.jl` between the time that I run `Pkg.update()` and the time that I create the pull request and everything will fall apart.
To me there are just too many balls in the air during this process to believe that there is a safe way of doing this.

simonbyrne:
I'm not sure, but I haven't had a problem so far. I do usually do `Pkg.update` before I `Pkg.tag`, but it actually doesn't matter that much (assuming of course that no one else has tagged a version of the same package), as all you're doing is creating new files (and not modifying old ones), so git is able to merge the changes easily.

PR #43, opened 2015-04-13
throw error when converting a factor to `DataArray`

randy3k:
Now, error is thrown when converting factor to `DataArray`. One have to convert it to `PooledDataArray`. The `DataFrame` method is also changed accordingly to use `DataArray` or `PooledDataArray` adaptively.


randy3k:
test passed.

PR #46, opened 2015-04-14
DataFrame to sexp conversion

randy3k:
As title

```julia
using RCall
using RDatasets
mtcars = dataset("datasets", "mtcars");
rmtcars = sexp(mtcars);
rprint(rmtcars)
```


simonbyrne:
nice!

PR #59, opened 2015-07-09
WIP: Major refactor to use RObject type.

simonbyrne:
There is still a bit more I would like to do, but I would appreciate some comments. This is a fairly major change, and will break some things, but I think it's a nicer interface.

R objects are now represented in 2 ways:
* behind-the-scenes as Ptr{S<:SEXPREC}
* user-facing RObject{S<:SEXPREC}, which provide GC-protected pointers

The advantage of this is that it is now clear whether or not an R object is protected from garbage collection.

In order to convert a Julia object to an R object:
```julia
RObject(x) # GC-protected
sexp(x)    # not protected
```
as a result, `sexp` is no longer exported. To do the reverse, use `rcopy`, with optional type:
```julia
rcopy(x) # default conversion
rcopy(T,x) # specific conversion
```

Other changes:
* SEXPRECs are now immutable
* whenever possible, the correct SEXPREC type is returned from functions, to improve type information
* nice wrappers to R functions (CAR, CDR, allocArray, setAttrib!, etc.)
* correctly call `Rf_xlength` for vectors.
* Use of `Base.vec` has been replaced by `unsafe_vec` (not exported)
* Use of `Base.names` has been replaced by `getNames` (exported)
* `protect` and `unprotect` are used to ensure Ptr{S<:SEXPREC} are not GC-ed.


randy3k:
Very promising :+1: .

src/functions.jl
```diff
@@ -1,29 +1,29 @@
...
-    l = preserve(sexp(ccall((:Rf_allocVector,libR),Ptr{Void},(Cint,Int),6,argn+1)))
-    s = l.p
-    ccall((:SETCAR,libR),Ptr{Void},(Ptr{Void},Ptr{Void}),s,f)
+    s = l = protect(allocArray(LangSxp,argn+1))
+    setcar!(s,sexp(f))
     for argv in args
-        s = ccall((:CDR,libR),Ptr{Void},(Ptr{Void},),s)
-        ccall((:SETCAR,libR),Ptr{Void},(Ptr{Void},Ptr{Void}),s,sexp(argv))
+        s = cdr(s)
+        setcar!(s,sexp(argv))
```
randy3k: Should we also protect `sexp(argv)`?

randy3k:
Just a naming issue. As far as I understand correctly, `SEXP`s are pointers to the `SEXPREC` but not subtypes of `SEXPREC`. It may make some sense to rename them as follows

```
@doc "R logical vector"->
type LglSxp_Struct <: SEXPREC     # type tag 10
    info::Cint
    attrib::Ptr{Void}
    p::Ptr{Void}
    pv::Ptr{Int32}
    length::Cint
    truelength::Cint
end

alias LglSxp Ptr{LglSxp_Struct}
```
How does it sound?

src/functions.jl
simonbyrne: From what I understand, I think it's fine as we're immediately binding it to a protected list. See, for example, the last block in [R-exts §5.11](http://cran.r-project.org/doc/manuals/r-patched/R-exts.html#Evaluating-R-expressions-from-C).

randy3k: Ya, you are probably right.

simonbyrne:
Good point. It does add more typs though, and the `RObject`s would then appear as `RObject{LglSxp_struct}`.

simonbyrne:
Their nomenclature is really confusing.

What about `LglSx`, or some variant? So we would have

```julia
abstract SExpr

immutable LglSx <: SExpr
...
end

typealias LglSxp Ptr{LglSx}
typealias Sxp Ptr{S<:SExpr}
```

randy3k:
If I understand correctly, SEXP or SXP stands for S expression. `LglSx` or `SExpr` may not be able to reflect the structure very precisely. 

How about

```julia
abstract SEXPREC

immutable LglSxpREC <: SEXPREC
...
end

typealias LglSxp Ptr{LglSxpREC}
typealias Sexp Ptr{S<:SEXPREC}
```

But I don't know what "REC" means...

simonbyrne:
> But I don't know what "REC" means...I searched over the source files but couldn't find any hints.

I would assume "Record"?

randy3k:
I found a presentation at userR 2004 http://www.r-project.org/conferences/useR-2004/Keynotes/Dalgaard.pdf
it says Symbolic EXPression RECord.

It makes sense to me to use `LglSxpRec` or `LglSxpREC`. And for the generic structure, `SEXPREC` or `SxpRec`

simonbyrne:
@randy3k I've renamed some things: let me know what you think.

randy3k:
As `sexp` returns SEXP rather than SEXPREC, should we define
`sexp(::Type{LglSxp},v)` instead of `sexp(::Type{LglSxpREC},v)`?

With that sense, I think we only have to export `Sxp` types but not the `SxpRec` types.

simonbyrne:
I see your point, but I'm reluctant for the following reasons:
* Users shouldn't have to care about the distinction between `Sxp` and `SxpRec`: this is only important for people using the lower level interface (basically, package developers who want to interact with R).
* The problem is that the `Sxp` types are just aliases, so the user will see `Ptr{FooSxpRec}`, which makes matters more confusing.
* From an aesthetic point of view `RObject{FooSxpRec}` is nicer than `RObject{Ptr{FooSxpRec}}`.
* Having an `SEXP` be a pointer is probably more due to limitations of C, rather than a semantic choice.
* We don't mirror the C interface in other ways: i.e. we use different structs for each type, `SEXP` is a (concrete) `Ptr{Void}`, whereas `Sxp` is an abstract type.

Having given this more thought, I was actually thinking it might be better to go back to using the original names, with the `Sxp` types being immutables, something like:
```
abstract Sxp # subtypes correspond to SEXPREC structs
typealias SxpPtr{S<:SxpRec} Ptr{S} # subtypes match SEXP arguments/

immutable LglSxp <: Sxp
...
end
typealias LglSxpPtr Ptr{LglSxp}
```
That way, users will see `RObject{LglSxp}`, which should be a bit nicer.

I realise this doesn't mirror the C interface, but I think as long as we document it, it shouldn't be a problem.

randy3k:
Agree. `LglSxpPtr`/`LglSxp` are more readable than ``LglSxp`/`LglSxpRec`.

simonbyrne:
@dmbates You mention in [your notes](https://github.com/JuliaStats/RCall.jl/blob/c1ff136864964cf2ac04b679f0c1b3b243df7e37/doc/MappingAPI.ipynb) that symbols are not garbage collected. Do you think this is likely to be the case for the foreseeable future? If so, we can special-case the `RObject{SymSxp}` to omit the `preserve` call.

dmbates:
@simonbyrne Yes, I expect that we will be able to count on symbols not being garbage collected for the foreseeable future.

simonbyrne:
Okay, now we have testing on all platforms. I was thinking of merging this in, cleaning up the docs and then tagging a new release. Any objections @randy3k @dmbates?

PR #60, opened 2015-07-12
use dframes.jl

randy3k:
@simonbyrne 

`using DataFrames` would source `dataframes.jl` on Mac OS. Have to change the filename to sth else.


simonbyrne:
Ah, good point.

PR #67, opened 2015-09-02
patch enable RCall can used in embedded julia 

armgong:
https://github.com/armgong/rcall (r package prototype) is R package embedded julia into R through RCall.jl

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -51,11 +51,13 @@ end
 
 
 function __init__()
-    argv = ["REmbeddedJulia","--silent","--no-save"]
-    i = ccall((:Rf_initEmbeddedR,libR),Cint,(Cint,Ptr{Ptr{UInt8}}),length(argv),argv)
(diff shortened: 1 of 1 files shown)


simonbyrne:
Rather than create a global variable in Main, it might be better to try to check directly if R is running. There doesn't seem to be an API to querythis, but based on this discussion:
https://stat.ethz.ch/pipermail/r-devel/2015-May/071069.html
you could try
```julia
R_inited = unsafe_load(cglobal((:R_NilValue,libR),Ptr{Void})) != C_NULL
```

armgong:
thanks,  i will try this tomorrow, and resummit patch.

armgong:
btw, after this pull request , rjulia can call RCall.jl no need modify,so I delete https://github.com/armgong/rcall

simonbyrne:
Nice! I'm super excited about having a 2-way interface.

Issue #49, opened 2015-04-15
interact with R event loops


simonbyrne:
In order to get features like graphics windows working correctly, we need to be able to interact with the R event loops. Some details are covered in [R-ext §8](http://cran.r-project.org/doc/manuals/r-release/R-exts.html#Linking-GUIs-and-other-front_002dends-to-R).

From what I understand, we need to implement (some of) the functionality provided by `run_Rmainloop`.

randy3k:
Eventloop was implemented in RCalling.jl and Rif.jl. See for example [this link](https://github.com/randy3k/RCalling.jl/blob/15e38fce6e803773642dc278877d3971234846c2/deps/init.c#L35). The code was written in C, however, it should be easy to port it to Julia. I can submit a PR this afternoon.

randy3k:
Please checkout the current master branch for the eventloop. Basically, there are two functions: `RCall.rgui_start()` and `RCall.rgui_stop()`.

simonbyrne:
Once we have callbacks sorted out, we could automate this using `setHook`.

randy3k:
In the [eventloop](https://github.com/randy3k/RCall.jl/blob/master/src/eventloop.jl#L3), there was a dirty fix related to `sigint`. The problem was when a `sigint` is signaled from Julia (pressing Control - C), the `R_interrupts_pending` is set to state 1. Without a proper R eventloop, `R_interrupts_pending` remains state 1. And it causes segfault when evaluating `R_ProcessEvents`. So `R_interrupts_pending` was forced to be 0 in there.

For a while, I only think that it is needed for `R_ProcessEvents`. Until just now, I discovered this when I pressed a `sigint` somewhere.

```julia
julia> using RCall

julia> reval("gc()")
RCall.RObject{RCall.RealSxp}
         used (Mb) gc trigger (Mb) max used (Mb)
Ncells 219236 11.8     460000 24.6   350000 18.7
Vcells 322397  2.5     786432  6.0   786432  6.0


julia> ^C

julia> reval("gc()")
ERROR: RCall.jl 

 in error at error.jl:22
 in reval_p at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:11
 in reval_p at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:19
 in reval at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:29
 in reval at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:31

[...]

PR #75, opened 2015-10-27
setHook for eventloop

randy3k:
It is an attempt to handle #49. In here, "plot.new" and "grid.newpage" hooks are being used to trigger the loop.

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -158,6 +158,9 @@ function __init__()
     # we can't use Rf_PrintWarnings as not exported on all platforms.
     rcall_p(:options,warn=1)
 
+    # R gui eventloop
+    !Rinited && isinteractive() && rgui_init()
+
     # IJulia hooks
     isdefined(Main, :IJulia) && Main.IJulia.inited && ijulia_init()
 end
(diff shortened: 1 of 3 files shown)


randy3k:
@simonbyrne any comments about the hooks?

simonbyrne:
Sorry I haven't had a chance to look at this yet: I'll try to get a look at it in the next few days.

simonbyrne:
This was roughly what I had in mind. 
* Wouldn't you want to make the call to `rgui_start` silent, as the hook can be invoked multiple times?
* Does `plot.new` not get called before `grid.newpage`?

randy3k:
The hook is called everytime a plot is made, and there are two major plotting devices, `graphics` and `grid`. `plot.new` is needed for R base, for example, `plot` and `hist`. And `grid.newpage` is needed for example, `ggplot2` which uses `grid`.

randy3k:
@simonbyrne Somehow Appveyor CI is stuck, could you restart it?

randy3k:
Look like Appveyor fails due to Dataframes.jl. I guess it is ready to merge.

PR #76, opened 2015-10-27
impove rprint function

randy3k:
`Rf_PrintValue` can cause segfault if S3 objects has custom print function as it doesn't use `R_tryEval`.


randy3k:
`Rf_PrintValueRec` not found on unix!?

simonbyrne:
> Rf_PrintValueRec not found on unix!?

I've been bitten by this a number of times: there is absolutely no consistency on what is exported across platforms. And if you ask on r-devel, the response is generally: "if it isn't mentioned in R-exts, we don't support it"

randy3k:
This is confusing. By the way, if one want to under how we need this, try this
```julia
reval("library(ggplot2)");
reval("ggplot()")
```
and this'll kill your julia session.

PR #77, opened 2015-11-20
add basic opearators

randy3k:
This allows us to do something like

```julia
using RCall
using RDatasets
mtcars = dataset("datasets", "mtcars");
@rlibrary ggplot2
p = ggplot(mtcars, aes_string(x="MPG", y="Disp")) + geom_line();
p
```

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -30,6 +30,7 @@ include("callback.jl")
 include("IJulia.jl")
 include("io.jl")
 include("rstr.jl")
+include("operators.jl")
 
 """
 R global Environment.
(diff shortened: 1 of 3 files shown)


nalimilan:
Maybe add support for all R operators while you're at it (for a list, see e.g. http://www.programiz.com/r-programming/operator)? This raises the question about how to represent `%*%` and similar, since they are not valid Julia syntax.

randy3k:
I also tried to override the short-circuit operators `&&` and `||`. It seems that these two operators are handled differently and cannot be overrided.

randy3k:
I guess we should merge the basics now and add more later.

Issue #87, opened 2016-01-27
Issue when converting back to Dict


ranjanan:
The following code segment
```
d = RObject(Dict(1=>2, 2=>3))
rcopy(Dict, d)
```
gives the following error:
```
ERROR: MethodError: `start` has no method matching start(::Type{Pair{Any,Any}})
 in rcopy at /Users/ranjan/.julia/v0.4/RCall/src/convert-base.jl:230
 in rcopy at /Users/ranjan/.julia/v0.4/RCall/src/convert-base.jl:20
```

ranjanan:
Pull request #88 fixes this.

PR #88, opened 2016-01-27
Fix dictionary conversion to Julia, add more tests

ranjanan:
Fix Dict issue

### DIFF
--- a/src/convert-base.jl
+++ b/src/convert-base.jl
@@ -71,7 +71,7 @@ sexp(::Type{CharSxp},sym::Symbol) = sexp(CharSxp,string(sym))
 
 rcopy{T<:AbstractString}(::Type{T},s::CharSxpPtr) = convert(T, bytestring(unsafe_vec(s)))
 rcopy(::Type{Symbol},s::CharSxpPtr) = symbol(rcopy(AbstractString,s))
-
+rcopy(::Type{Int}, s::CharSxpPtr) = parse(Int, rcopy(s))
 
 "Create a `StrSxp` from an `AbstractString`"
 sexp(::Type{StrSxp}, s::CharSxpPtr) =
@@ -227,7 +227,8 @@ sexp(d::Associative) = sexp(VecSxp,d)
 
 function rcopy{A<:Associative,S<:VectorSxp}(::Type{A}, s::Ptr{S})
     a = A()
-    K,V = eltype(a)
+    K = keytype(a)
+    V = valtype(a)
     for (k,v) in zip(getNames(s),s)
         a[rcopy(K,k)] = rcopy(V,v)
     end
@@ -237,7 +238,8 @@ end
 function rcopy{A<:Associative,S<:PairListSxp}(::Type{A}, s::Ptr{S})
     protect(s)
     a = A()
-    K,V = eltype(a)
+    K = keytype(a)
+    V = valtype(a)
     for (k,v) in s
         a[rcopy(K,k)] = rcopy(V,v)
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,4 +1,4 @@
-using Base.Test
+using Base.Test, Compat
 hd = homedir()
 pd = Pkg.dir()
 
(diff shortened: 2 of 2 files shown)


ranjanan:
This fixes #87

src/convert-base.jl
```diff
@@ -227,7 +227,7 @@ sexp(d::Associative) = sexp(VecSxp,d)
 
 function rcopy{A<:Associative,S<:VectorSxp}(::Type{A}, s::Ptr{S})
     a = A()
-    K,V = eltype(a)
+    K,V = eltype(a).parameters
```
simonbyrne: I don't think we should be using these internal fields. Better to use:
```
K = keytype(a)
V = valtype(a)
```

Other than that, this looks great, thanks!

ranjanan:
@simonbyrne Incorporated the changes you've suggested.

simonbyrne:
Fantastic, thanks.

Actually, would you mind adding a test as well?

ranjanan:
Ah, I've written a bunch of tests. I thought about opening up another pull request for that. Shall I just add them all to this one?

simonbyrne:
Fantastic. Whatever is easier.

ranjanan:
There seems to be a problem with `rcopy(Cint(1))` . It says `Type cannot be constructed`. This seems to work on my Mac on Julia v0.4.1

simonbyrne:
This is on 0.3, which doesn't support calling bitstypes (such as `Cint(1)`). Add `@compat` outside (e.g. `@compat(Cint(1))`.

Perhaps time has come to deprecate 0.3, though that should be a separate PR.

ranjanan:
I'm afraid `@compat(Cint(1))` doesn't work. It still says `Type cannot be constructed`

simonbyrne:
Hmm, try `@compat(Int32(1))`.

ranjanan:
Oh, that works. Interesting to note that `@compat(Cint(1))` doesn't work even though they're just type-aliases.

simonbyrne:
Macro trickery I guess.

src/convert-base.jl
```diff
@@ -71,7 +71,7 @@ sexp(::Type{CharSxp},sym::Symbol) = sexp(CharSxp,string(sym))
 
 rcopy{T<:AbstractString}(::Type{T},s::CharSxpPtr) = convert(T, bytestring(unsafe_vec(s)))
 rcopy(::Type{Symbol},s::CharSxpPtr) = symbol(rcopy(AbstractString,s))
-
+rcopy(::Type{Int64}, s::CharSxpPtr) = parse(Int, rcopy(s))
```
simonbyrne: This should be `rcopy(::Type{Int}, ....`

ranjanan:
Oh I see. How will changing it to `Int` resolve it? Isn't `Int` typealias for `Int64` on Windows too? Should I also write a method for `Int32`s as well in that case?

nalimilan:
When you'll squash the commits before merging, it would be nice to use a more descriptive title than "update convert-base.jl".

ranjanan:
Oh, I haven't squashed before. Would `git rebase -i HEAD~5` squash all of my commits together?

nalimilan:
Yes, if you change the first word of the lines to `squash` (except for the commits you want to keep of course, whose message can be edited by choosing `edit`). Note that you can keep several commits as long as each of them is meaningful and passes the tests.

simonbyrne:
The problem was only on 32-bit Windows (`Int` aliases to `Int32` on 32-bit systems).

You can also use `git rebase -i master` to rebase from where you diverged from `master`.

ranjanan:
I had to merge a particular commit that was done before I last pulled. I hope that's okay? Or should I merge the two above too?

nalimilan:
You should never have to merge branches: rebase them instead, so that all commits make logical sense. Anyway, the last commit doesn't include any changes, so you should be able to squash it just like the others.

ranjanan:
@nalimilan I should be, but I get this error on `git rebase -i HEAD~2` : `error: could not apply 5e2a5ba... Update convert-base.jl`. Not quite sure why, any ideas?

simonbyrne:
Hmm, you seem to have commits on two different branches that have diverged from `origin/master`.

ranjanan:
Not quite sure why that is. I think I must have committed all these fixes without pulling the latest code in the first place. What that be the reason? And since Travis and Appveyor are happy, could we merge it? Or would the squash have to be done mandatorily?

nalimilan:
Doesn't sound great to pollute the commit log with partial commits like this. A radical solution is to do `git diff master > branch.diff` to get a patch file for the whole branch. Then you can do `git checkout -b backup` to create a copy of the branch (just in case), `git reset --hard master` to destroy all of your changes, and `git apply branch.diff` to apply the patch. Then you can commit it as you do usually.

simonbyrne:
I've rebased and pushed it manually. Thanks for sorting this out!

ranjanan:
Alright thanks! That fixes #87 then.

PR #90, opened 2016-01-29
Add docs to some exported functions

ranjanan:


dmbates:
@ranjanan Do you know why the tests are failing?  It seems to be a line where you have
```julia
@doc "doc string" functionName
```

The other uses of `@doc` seem to have the -> at the end of the docstring, which may be because of
```julia
using Docile
```

when testing against version 0.3.x

Perhaps a more relevant question is whether we want to continue to support 0.3 versions.

ranjanan:
@dmbates Apologies for delay, was travelling. 
>Do you know why the tests are failing? It seems to be a line where you have
>@doc "doc string" functionName.

Yes, that is correct.  Changing this to 
```
"Docs"
function 
```
should preserve compatibility with v0.3
>Perhaps a more relevant question is whether we want to continue to support 0.3 versions.

It doesn't seem overtly difficult to maintain backward compatibility in this package, unless you guys are planning an overhaul. @ViralBShah , is it desirable for these packages to continue supporting v0.3?

ViralBShah:
What does @simonbyrne and @tkelman have to say on 0.3?

tkelman:
Make an effort not to tag a new version that says it supports 0.3 but is actually broken there. It's fine to raise thr minimum version in require to 0.4, then bump the minor package version in the first tag that drops support for 0.3. Existing tags will stay as is (working, hopefully?), and new features will be 0.4 only.

ranjanan:
@simonbyrne Can we merge this one now though?

src/library.jl
```diff
@@ -25,6 +25,7 @@ function rwrap(pkg::ASCIIString,s::Symbol)
     m
 end
 
+"Import R Packages into Julia with or without alias. You can also use classic Python syntax to do this: `import *module-name* as *shorthand*`"
```
simonbyrne: should be `@rimport *module-name* as *shorthand*`

tkelman: seems a little strange to be referring to "classic python syntax" in a package to call R from Julia...

simonbyrne:
Thanks.

PR #93, opened 2016-03-01
WIP: Enable precompilation, bump Julia version requirement to 0.4

simonbyrne:
This modifies the internals to allow for package precompilation. Due to some of the necessary changes (e.g. we can no longer use `const` pointers), this is a breaking change, though hopefully the "high-level" interface should still work.

This also changes the library search functionality so that we no longer need to use `Pkg.build`.


simonbyrne:
Okay, I'm fairly happy with this (the AppVeyor failure is unrelated).

@randy3k any thoughts?

randy3k:
It is a great step. Thanks @simonbyrne.

I don't mind merging this to the master.
As I don't have much experience with the pre complication thing.
I am extremely busy this week, will check it out this weekend.

simonbyrne:
I've tagged a last version for 0.3, so I'll merge this soon.

Issue #82, opened 2016-01-15
Use of an R_str, or similar, macro?


dmbates:
I had been wondering for some time how a string modifier like the `'r`' in `r"my\w+regexp"`was defined and whether we could define `R"x <- seq(0,1,101)"` to mean `reval("x <- seq(0,1,101)")`, say.  It turns out that the magic is defining a macro with the name `R_str`.

Does it seem worthwhile allowing this kind of simplification of calls to `reval(str)` or, perhaps, `rcopy(reval(str))`.  I find that when using the `RCall` package I get tired of writing that sequence of calls over and over.

nalimilan:
Would an `@R` macro work? I suspect many R expressions are also valid Julia syntax, and that way one would benefit from syntax highlighting. Though it certainly makes sense to have a non-standard string literal, in particular if one wants to write several lines of R code using `"""`.

randy3k:
> I suspect many R expressions are also valid Julia syntax, and....

 I can list a number of exceptions

1. list subset `list$abc` and `list[["abc"]]`
2. variable names containing dots, e.g., `obj.num1`
3.  `%*%`, `%%`, `%in%`, etc and user defined operators.
4. assignment `<-` and `<<-`

simonbyrne:
I actually made a start on this on the [`rstr` branch](https://github.com/JuliaStats/RCall.jl/compare/rstr), currently unfinished. The idea was to support `$` substitution whenever the `$` was invalid in R, so that something like
```
R"sum(a$b)"
```
would be normal R syntax (looking up the `b` field of `a`, but
```
R"sum($b)"
```
would insert the Julia object `b` into the expression.

dmbates:
Any further progress, Simon?  I am running a workshop on "Julia for Statisticians" where RCall is extremely useful and would like to give a general prescription of

R"""
# your R code here
"""

simonbyrne:
Not yet. I'm looking at some other RCall improvements, so I might try to see what I can do in the next few days.

dmbates:
@simonbyrne Let me know if I can help in some way.

Would it be worthwhile dropping support for v0.3 in addition to other changes?  That could simplify the inline documentation and allow for removal of some of the calls to `@compat`

simonbyrne:
I was going to do that. I also want to get precompilation working.

PR #94, opened 2016-03-10
R_str macro: allows inlining R code directly into Julia, supporting limited variable substitution.

simonbyrne:
See #82.

This is somewhat limited: it only supports ASCII (due to https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16524), and you can't assign to Julia variables (getting that logic correct is a bit complicated). Nevertheless, it seems to work for simple snippets:

```julia
iris = rcopy(:iris)
model =  R"lm(Sepal.Length ~ Sepal.Width,data=$iris)"
```

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -12,7 +12,7 @@ export RObject,
    globalEnv,
    rcopy, rparse, rprint, reval, rcall, rlang,
    isNA, anyNA, isFactor, isOrdered,
-   @rimport, @rusing, @rput, @rget, @var_str
+   @rimport, @rusing, @rput, @rget, @var_str, @R_str
 
 
 include("setup.jl")
@@ -29,6 +29,7 @@ include("eventloop.jl")
 include("callback.jl")
 include("IJulia.jl")
 include("io.jl")
+include("rstr.jl")
 
 
 const globalEnv = Const.GlobalEnv
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -374,7 +374,7 @@ function getindex(e::Ptr{EnvSxp},s::Ptr{SymSxp})
     v == sexp(Const.UnboundValue) && error("$s is not defined in the environment")
     sexp(v)
 end
-getindex(e::Ptr{EnvSxp},s) = getindex(e,sexp(s))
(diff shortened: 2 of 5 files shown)


dmbates:
Thanks very much for doing this, Simon.  I have been looking forward to having this capability.

Does the logic distinguish between the R use of a $ symbol, as in Formaldehyde$optden, and its use for variable substitution?

simonbyrne:
Yes: the logic is basically that it tries to parse it as an R expression: if that fails, then it tries to variable substitution via a Julia expression. So
```julia
R"sum($iris$Sepal.Length)"
```
will substitute in the Julia `iris` variable, but then since the rest is valid R syntax, the rest is interpreted as it would be in R.

nalimilan:
Isn't that a bit too clever? I would have disabled interpolation altogether to make things simpler.

simonbyrne:
@nalimilan But that wouldn't buy us anything over `reval`, other than escaping `$`.

I wanted to be able to do things like
```julia
X = ...
R"plot($X)"
```

simonbyrne:
It's somewhat lucky that most cases are not ambiguous: `$` is basically only valid R syntax when followed directly by a symbol (e.g. `aa$bb`), for which it wouldn't make sense to do substitution.

dmbates:
The changes look good to me.  I think we should go ahead and merge them into master for more extensive testing by users.  I would have been happy with an `R_str` macro that simply escaped the `$` characters and piped the result to `reval` but Simon's approach has a lot more going for it.

@simonbyrne In your comment about `$` in valid R syntax I think "followed directly by a symbol" is not enough.  Look at the example in your previous comment, `$X` is `$` followed by a symbol.  Almost all the usage in R is `symbol` `$` `symbol` but `R expr` `$` `symbol` can also be used.

simonbyrne:
You're right, that's a mistake. I meant `$` is only valid R syntax when directly *following* a symbol, but as you point out that isn't correct. Even then, I can't think of any cases where the usage would be ambiguous, since `expr symbol` isn't valid syntax.

simonbyrne:
Okay, I've merged into master. I won't tag a new version for a bit until we've ironed out the kinks.

Issue #109, opened 2016-05-07
GC: Sometimes there is a warning when the R macro and rcopy are using inside a loop


diegozea:
i.e.

```julia
julia> using RCall

julia> for i in 1:10
           n = rand(1:1000)
           ie = Symmetric(rand(n,n))
           rcopy(R"""hclust(as.dist($ie), method="average")""")[:order]
       end

julia> for i in 1:10
           n = rand(1:1000)
           ie = Symmetric(rand(n,n))
           rcopy(R"""hclust(as.dist($ie), method="average")""")[:order]
       end
WARNING: RCall.jl Warning in rm("##RCall##7259") : objeto '##RCall##7259' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7259") : objeto '##RCall##7259' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7259") : objeto '##RCall##7259' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7085") : objeto '##RCall##7085' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7259") : objeto '##RCall##7259' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7085") : objeto '##RCall##7085' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7085") : objeto '##RCall##7085' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
WARNING: RCall.jl Warning in rm("##RCall##7085") : objeto '##RCall##7085' no encontradoerror in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")

julia> for i in 1:10
           n = rand(1:1000)
           ie = Symmetric(rand(n,n))
           R"""hclust(as.dist($ie), method="average")"""
       end
[...]

Issue #107, opened 2016-04-29
Error transferring vectors


davidanthoff:
I'm not sure whether the title is a good error description. I'm trying this:

````
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.4.5 (2016-03-18 00:58 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-w64-mingw32

julia> using RCall

julia> R"library(ggplot2)";

julia> x = randn(100);

julia> y = randn(100);

julia> R"qplot($x,$y)"
RCall.RObject{RCall.VecSxp}
Error showing value of type RCall.RObject{RCall.VecSxp}:
ERROR: RCall.jl Error in eval(expr, envir, enclos) : object '##RCall##1' not found

 in error at error.jl:22
````
Shouldn't that work?

**Comment by randy3k** [comment]
It is due to a bug in RCall, the following should work.
```julia
R"print(qplot($x,$y))"
```

@simonbyrn
It seems that this issue affects all S3 objects with custom print method. `@R_str` has removed the variable before the `rprint` command is invoked.

randy3k:
Look like we need to preserve the variables `##RCall##{number}` until the result returned by `@R_str` has been gc'ed.

randy3k:
the previous fix is not correct, reopen the issue.

randy3k:
#110 should fix it. Please let us know if problem persists.

davidanthoff:
It all works with the latest master, thanks a lot!

One minor thing: the axis labels now are ``$x`` and ``$``, i.e. they include the escape character. Might be nicer if the R variables used just had the julia name variables, without the ``$``?

randy3k:
I think the reason of such implementation is to differentiate R expressions and Julia expressions.

It is possible to pass Julia expressions in `R_str`.
```julia
x = rand(10); # "x" in Julia
globalEnv[:x] = rand(10); # "x" in R
R"plot(x, $x)";
```

simonbyrne:
[...]

PR #110, opened 2016-05-07
Evaluate R string macro inside own environment.

simonbyrne:
This means that we don't need to bother cleaning up after, hence fixing #107 and #109. Also, changes symbol names to be those of the underlying Julia variables, prefixed with `$`, which makes plots look nicer.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,5 @@
 julia 0.4
+DataStructures
 DataArrays
 DataFrames
 @windows WinReg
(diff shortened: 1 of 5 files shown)


randy3k:
Though it is being noted that the expressions are evaluated in a new environment so that `<<-` should be used for assigning variables, I believe it could be a major confusion for new users. They may expect somethings like the following would work.

```julia
julia> R"x <- 1"
RCall.RObject{RCall.RealSxp}
[1] 1

julia> R"x"
ERROR: RCall.jl Error: object 'x' not found
```

simonbyrne:
That's why i was initially hesitant to do it, but now I've come to think
it's probably worth it.

PR #119, opened 2016-06-18
R REPL mode #115

### LINKED ISSUE #115: REPL mode
Would be really cool if there was a REPL mode that essentially wrapped the ``R"..."`` macro.

randy3k:
type `$` (key suggested by Douglas) to activate repl mode and type <kbd>backspace</kbd> to leave r mode

```
R> x=1

R> x
[1] 1
```

close #115

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -34,6 +34,8 @@ include("callback.jl")
 include("IJulia.jl")
 include("rstr.jl")
 include("operators.jl")
+include("repl.jl")
+
 
 """
 R global Environment.
--- a/src/iface.jl
+++ b/src/iface.jl
@@ -2,13 +2,17 @@
 Evaluate an R symbol or language object (i.e. a function call) in an R
 try/catch block, returning a Sxp pointer.
 """
-function reval_p{S<:Sxp}(expr::Ptr{S}, env::Ptr{EnvSxp})
+function reval_p{S<:Sxp}(expr::Ptr{S}, env::Ptr{EnvSxp}, silent::Bool=false)
     err = Array(Cint,1)
     protect(expr)
     protect(env)
     val = ccall((:R_tryEval,libR),UnknownSxpPtr,(Ptr{S},Ptr{EnvSxp},Ptr{Cint}),expr,env,err)
     unprotect(2)
-    print(STDOUT, takebuf_string(printBuffer))
+    if !silent && nb_available(printBuffer) != 0
+        # it is needed when the expression contains `print` or `cat` statements
+        # which are not returning results
+        print(STDOUT, takebuf_string(printBuffer))
+    end
     if err[1] !=0
         error("RCall.jl ", takebuf_string(errorBuffer))
     elseif nb_available(errorBuffer) != 0
@@ -17,12 +21,12 @@ function reval_p{S<:Sxp}(expr::Ptr{S}, env::Ptr{EnvSxp})
     sexp(val)
 end
 
-function reval_p(expr::Ptr{ExprSxp}, env::Ptr{EnvSxp})
+function reval_p(expr::Ptr{ExprSxp}, env::Ptr{EnvSxp}, silent::Bool=false)
     local val           # the value of the last expression is returned
     protect(expr)
     try
         for e in expr
-            val = reval_p(e,env)
+            val = reval_p(e,env, silent)
         end
     finally
         unprotect(1)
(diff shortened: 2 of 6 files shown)


davidanthoff:
Also works on Windows. Very nice! But so far string splatting is not enabled, right? So I can't for example have julia variables splatted in via ``$x`` etc.?

randy3k:
There was an [issue](https://github.com/JuliaStats/RCall.jl/pull/110) of handling the memory of transferred variables and now each `@R_str` is evaluated in a separated environment. Until we figure out the best solution to that issue, `$x` would be not possible.

I guess we could add some helper functions in the R mode to help copying variables from and to Julia.

davidanthoff:
Hm, I do think that the real value of a REPL would come from the ability to use the splatting stuff, so if that could be enabled somehow it would be great.

armgong:
the patch seem great ,but maybe need code completion when user press TAB key, and R code completion in  src/library/utils/R/completion.R and src/unix/sys-std.c on linux.

armgong:
try add a function make parse incomplete works ,hope this useful .
but sorry currently I don't have git and diff ,so just paste function, and we still need more code when user paste code to LineEdit (like julia does)
~~~julia
function return_callback(s)
        st = sexp(String(LineEdit.buffer(s)))
        protect(st)
        status = Array(Cint,1)
        expr = ccall((:R_ParseVector,libR),UnknownSxpPtr,
                    (Ptr{StrSxp},Cint,Ptr{Cint},UnknownSxpPtr),
                    st,-1,status,sexp(Const.NilValue))
        unprotect(1)
        s = status[1]
        if s != 1
          return false
        else
          return true
        end

end

function repl_init()
    repl = Base.active_repl
    mirepl = isdefined(repl,:mi) ? repl.mi : repl

    main_mode = mirepl.interface.modes[1]

    panel = LineEdit.Prompt("R> ";
        prompt_prefix=Base.text_colors[:blue],
        prompt_suffix=main_mode.prompt_suffix,
        on_enter=return_callback)
~~~

davidanthoff:
Just to come back to the splatting question: my main use case for all of this right now is plotting with ggplot2. If I had to manually transfer stuff from Julia to R and then could just use the R REPL to issue the plotting command, I would probably just stick with the R macro so that I can use the splatting stuff.

I think I saw a comment on some other issue by @simonbyrne that only splatter vars could be put in their own R environment for the R string macro, but the rest executed in the global context. That could also work for the REPL, right?

dmbates:
The `<` character is already in use for switching the REPL context in the `Cxx` package.  It would be best to choose another character, perhaps `$` which already has an unorthodox meaning in R.

randy3k:
> I think I saw a comment on some other issue by @simonbyrne that only splatter vars could be put in their own R environment for the R string macro, but the rest executed in the global context. That could also work for the REPL, right?

Actually, the whole chunk of code is evaluated in its own environment. (I am on the master branch, older code is different).

```julia
julia> x = 1
1

julia> R"foo = $x"
RCall.RObject{RCall.IntSxp}
[1] 1


julia> R"foo"
ERROR: RCall.jl Error: object 'foo' not found

 in error at error.jl:22
 in reval_p at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:17
 in reval_p at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:29
 in reval at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:43
 in reval at /Users/Randy/.julia/v0.4/RCall/src/iface.jl:44
```

The reason of this implementation is because we don't know how long `$x` should be kept in the memory of R. Certainly we don't want to erase it immediately or to keep it indefinitely. 

In the current implementation, the whole environment will be gc'ed (in the future) if the returning object of the `@R_str` macro is no longer referenced in julia.

randy3k:
I have made the modification suggested by Douglas.

davidanthoff:
> The reason of this implementation is because we don't know how long $x should be kept in the memory of R. Certainly we don't want to erase it immediately or to keep it indefinitely.

Hm, that seems pretty simple to me: it should be kept in memory exactly for the duration of one ``R""`` call or one REPL command, and no longer. If I want to keep it in R memory for longer, I'll have to assign it to an R variable, which would be preserved across different ``R""`` evaluations or REPL commands.

So shouldn't the implementation for the REPL and the ``R""`` macro first of all be the same, and second be just such that everything is evaled in the global R environment, and that anything splatted into it is put into a separate R environment that can be freed after the string macro call or REPL call?

I can just promise you that if you can demo plotting a DataFrame with ggplot2 from the R REPL via splatting, people will be ecstatic. It would simply be the perfect workflow: do some analytics in julia, and then plot the stuff with one simple line of code from the R REPL...

randy3k:
> Hm, that seems pretty simple to me: it should be kept in memory exactly for the duration of one ``R""`` call or one REPL command, and no longer. If I want to keep it in R memory for longer, I'll have to assign it to an R variable, which would be preserved across different ``R""`` evaluations or REPL commands.

It was exactly what our old code is doing. However, in situation like ggplot2, `ggplot` creates the plots at "show" time but not the evaluation time. `ggplot` would not work if the variables are removed immediately after the R marco. It was also why you filed the issue https://github.com/JuliaStats/RCall.jl/issues/107. A dirty solution is to remove the variables after show time, but it comes with another problem, what if a user do `R"ggplot(...)";` with the semicolon? There is no "show" time in such case.

> So shouldn't the implementation for the REPL and the ``R""`` macro first of all be the same, and second be just such that everything is evaled in the global R environment, and that anything splatted into it is put into a separate R environment that can be freed after the string macro call or REPL call?

The REPL call is a more continuous flow than the R macro, there are some subtle differences between them. For example, in Julia, the semi colon will suppress the "show" time, however, such a thing doesn't exist in R. To avoid printing long output, we have the check the `R_Visible`.

```julia
julia> R"x<-1";

julia> R"x<-1"
RCall.RObject{RCall.RealSxp}
[1] 1
```

> I can just promise you that if you can demo plotting a DataFrame with ggplot2 from the R REPL via splatting, people will be ecstatic. It would simply be the perfect workflow: do some analytics in julia, and then plot the stuff with one simple line of code from the R REPL...

I agree with you very much, ggplot2 is also the reason I use RCall

randy3k:
@armgong 
Just added tab completion (inspired by https://github.com/IRkernel/IRkernel/blob/a0818319d3b91be54789fdc181879a3d0189e6fa/R/kernel.r#L236)

PS: I found that the `utils` completions is more powerful than the terminal console completion.

armgong:
thank you for tab completion ,but one more question, how to return back to Julia REPL?

randy3k:
<backspace>

davidanthoff:
Would be good to add docs for this, and make it a practice to document new features.

PR #124, opened 2016-06-20
evaluating Rstr in globalEnv

randy3k:
Improvement to handle the variable substitution, but it seems that there is an memory leak issue. https://github.com/JuliaStats/RCall.jl/issues/123

### DIFF
--- a/src/rstr.jl
+++ b/src/rstr.jl
@@ -66,17 +66,16 @@ function rscript(script::Compat.ASCIIString)
             expr,i = parse(script,i,greedy=false)
 
             if isa(expr,Symbol)
-                sym = "\$$expr"
+                sym = "$expr"
             else
-                sym = "\$($expr)"
+                sym = "($expr)"
                 # if an expression has already appeared, we generate a new symbol so it will be evaluated twice (e.g. `R"$(rand(10)) == $(rand(10))"`)
                 if haskey(symdict, sym)
                     sym *= "##$k"
                 end
             end
             symdict[sym] = expr
-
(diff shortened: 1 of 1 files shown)


randy3k:
I guess I have figured out the memory leak issue. 

Consider
```julia
macro foo()
    quote
        ret = reval("1:10^8")
        ret
    end
end
```

```
julia> @foo;

julia> gc()

R> gc()
           used  (Mb) gc trigger  (Mb) max used  (Mb)
Ncells   218360  11.7     460000  24.6   350000  18.7
Vcells 50323035 384.0   72867244 556.0 50324576 384.0
```

randy3k:
I have posted an issue on Julia https://github.com/JuliaLang/julia/issues/17044.

randy3k:
@simonbyrne I have added the protect and unprotect pair for `env`.

randy3k:
It turns out we should use a `let` block to prevent variables to be exported to global env.

src/rstr.jl
```diff
@@ -103,24 +102,27 @@ All such Julia expressions are evaluated once, before the R expression is evalua
...
         push!(blk_ld.args,:(env[$rsym] = $(esc(expr))))
     end
     quote
-        env = newEnvironment()
-        protect(env)
-        $blk_ld
-        ret = reval($script, env)
-        unprotect(1)
-        ret
+        let
```
yuyichao:
```jl
let ret, env = protect(newEnvironment())
    .....
end
```

randy3k: thanks

PR #126, opened 2016-06-21
substituite julia variables in REPL mode

### LINKED ISSUE #121: Enable variable splatting support in REPL
I am aware that this is tricky, but maybe this issue here can serve as a discussion place on how to achieve this.

@simonbyrne wrote this in #117:
> One idea that has occurred to me is that the R"" macro could evaluate in the global environment, but the substituted variables could be in a macro-specific environment.

Maybe that is a starting point?

randy3k:
close #121 

@davidanthoff, please test it.


davidanthoff:
Seems to work, so cool!! Thanks a ton for the quick turnaround!

PR #128, opened 2016-06-22
only keep R internals as camelCase functions

randy3k:
follow julia's convention to name functions with underscores.


simonbyrne:
I'm not so keen about this: while it's probably a good idea for exported functions, a lot of the internal ones (like `allocArray`) are so named because they're wrappers of the C function of the same name. I find that keeping the same name does make it easier to keep track of what is going on.

randy3k:
@simonbyrne thanks for the comment. After some thought, it seems a better idea to keep the internals camel cased. A quick check reveals that RStudio has similar coding style.

randy3k:
@simonbyrne have a check. For now, all camel cased functions are direct wrapper of the R C API, I just realize having camel cases in those functions might alert us that those functions may not be "safe".

randy3k:
I'm going to merge it if there are no other comments.

PR #145, opened 2016-09-26
Update to use NullableArrays and CategoricalArrays

randy3k:
TODO

- [x] Handle ordered factor
- [x] Optionally convert scalar NA to Nullable instead of NullableArray

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -1,11 +1,11 @@
 __precompile__()
 module RCall
-using Compat, DataFrames, DataArrays
+using Compat, DataFrames, NullableArrays, CategoricalArrays
 import Compat.String
 
 import DataStructures: OrderedDict
 
-import Base: eltype, show, convert, isascii,
+import Base: eltype, show, convert, isascii, isnull,
     length, size, getindex, setindex!, start, next, done, names
 
 export RObject,
@@ -13,7 +13,7 @@ export RObject,
    ListSxp, VecSxp, EnvSxp, LangSxp, ClosSxp, S4Sxp,
    getattrib, setattrib!, getnames, setnames!, getclass, setclass!, attributes,
    globalEnv,
-   isna, anyna, isnull,
+   isna, anyna,
    rcopy, rparse, rprint, reval, rcall, rlang,
    @rimport, @rlibrary, @rput, @rget, @var_str, @R_str
 
--- a/src/convert-data.jl
+++ b/src/convert-data.jl
@@ -11,17 +11,23 @@ function rcopy(::Type{NullableArray}, s::Ptr{IntSxp})
     isFactor(s) && error("$s is a R factor")
     NullableArray(rcopy(Array,s), isna(s))
 end
+function rcopy(::Type{CategoricalArray}, s::Ptr{IntSxp})
+    isFactor(s) || error("$s is not a R factor")
+    refs = UInt32[x for x in s]
+    levels = rcopy(getattrib(s,Const.LevelsSymbol))
+    pool = CategoricalPool(levels, isOrdered(s))
+    CategoricalArray(refs, pool)
+end
 function rcopy(::Type{NullableCategoricalArray}, s::Ptr{IntSxp})
     isFactor(s) || error("$s is not a R factor")
     refs = UInt32[isna(x) ? zero(UInt32) : UInt32(x) for x in s]
-    pool = CategoricalPool(rcopy(getattrib(s,Const.LevelsSymbol)))
+    levels = rcopy(getattrib(s,Const.LevelsSymbol))
+    pool = CategoricalPool(levels, isOrdered(s))
     NullableCategoricalArray(refs, pool)
 end
-
 function rcopy(::Type{DataFrame}, s::Ptr{VecSxp})
     isFrame(s) || error("s is not a R data frame")
-    DataFrame(Any[isFactor(c)? rcopy(NullableCategoricalArray, c) : rcopy(NullableArray, c) for c in s],
-              rcopy(Array{Symbol},getnames(s)))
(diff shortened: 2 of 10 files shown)


src/RCall.jl
```diff
@@ -1,19 +1,19 @@
 __precompile__()
 module RCall
-using Compat, DataFrames, DataArrays
+using Compat, DataFrames, NullableArrays, CategoricalArrays
 import Compat.String
 
 import DataStructures: OrderedDict
 
-import Base: eltype, show, convert, isascii,
+import Base: eltype, show, convert, isascii, isnull,
```
nalimilan: You shouldn't need `import` for this, merely `using` AFAICT.

test/dataframe.jl
```diff
@@ -1,31 +1,52 @@
-using DataArrays,DataFrames
+using NullableArrays,CategoricalArrays,DataFrames
 
-v110 = rcopy(DataArray,reval("x <- 1:10"))
-@test isa(v110,DataVector)
-@test eltype(v110) == Cint
+@test rcopy(Nullable, RObject(1)).value == 1
+@test rcopy(Nullable, RObject("abc")).value == "abc"
+@test rcopy(RObject(Nullable(1))) == 1
+@test rcopy(Nullable, RObject(Nullable(1, true))).isnull
```
nalimilan: You should really avoid accessing the `isnull` field directly, as it's likely to go away soon (https://github.com/JuliaLang/julia/pull/18510). Use the accessor instead. (This also applies in other places.)

test/dataframe.jl
```diff
@@ -1,31 +1,52 @@
...
+@test isa(station,NullableCategoricalArray)
 
-@test rcopy(DataArray,"c(NA,TRUE)").na == @data([NA,true]).na
-@test rcopy(DataArray,"c(NA,1)").na == @data([NA,1.0]).na
-@test rcopy(DataArray,"c(NA,1+0i)").na == @data([NA,1.0+0.0*im]).na
-@test rcopy(DataArray,"c(NA,1L)").na == @data([NA,one(Int32)]).na
-@test rcopy(DataArray,"c(NA,'NA')").na == @data([NA,"NA"]).na
-@test_throws ErrorException rcopy(DataArray,"as.factor(c('a','a','c'))")
-@test rcopy(PooledDataArray,"as.factor(c('a','a','c'))").pool == ["a","c"]
+@test rcopy(NullableArray,"c(NA,TRUE)").isnull == NullableArray([true,true], [true,false]).isnull
```
nalimilan: Same here for `NullableArray`. You can use `isequal` to test for equality of both contents and missingness, which is even better.

src/convert-data.jl
```diff
@@ -39,12 +68,28 @@ function sexp(v::DataArray)
...
-    setattrib!(rv, Const.LevelsSymbol, sexp(v.pool))
-    setattrib!(rv, Const.ClassSymbol, sexp("factor"))
-    rv
+## CategoricalArray to sexp conversion.
+for typ in [:NullableCategoricalArray, :CategoricalArray]
+    @eval begin
+        function sexp{T<:Compat.String,N,R<:Integer}(v::$typ{T,N,R})
+            rv = protect(sexp(v.refs))
+            try
+                for (i,isna) = enumerate(v.refs .== 0)
```
nalimilan: `v.refs .== 0` will allocate a vector. Better do `== 0` inside the loop.

src/convert-data.jl
```diff
@@ -39,12 +68,28 @@ function sexp(v::DataArray)
...
+                    if isna
+                        rv[i] = naeltype(eltype(rv))
+                    end
+                end
+                setattrib!(rv, Const.LevelsSymbol, sexp(v.pool.index))
```
nalimilan: Please use `CategoricalArrays.index(v)`, `ordered(v)` and `levels(v)`. Private fields should really not be accessed directly as they could change.

src/RCall.jl
randy3k: it is because RCall also has a isnull method.

src/convert-data.jl
randy3k: `CategoricalArrays.index` does not work?

```julia
julia> v = CategoricalArray(repeat(["a", "b"], inner = 5));

julia> CategoricalArrays.levels(v)
2-element Array{String,1}:
 "a"
 "b"

julia> CategoricalArrays.index(v)
ERROR: MethodError: no method matching index(::CategoricalArrays.CategoricalArray{String,1,UInt32})
Closest candidates are:
  index(::CategoricalArrays.CategoricalPool{T,R<:Integer,V}) at /Users/Randy/.julia/v0.5/CategoricalArrays/src/pool.jl:161
```

src/convert-data.jl
randy3k: thanks for catching it

test/dataframe.jl
nalimilan: Yes, you need to do `NullableArray(1:3)`.

src/convert-data.jl
nalimilan: Ah, indeed, at first I didn't intend it to be used outside of the package. For now the best approach is `CategoricalArrays.index(v.pool)`. I'll try to see if I can improve the API in next releases.

test/dataframe.jl
randy3k: `isequal(NullableArray(1:3), 1:3)` or `isequal(NullableArray(1:3), collect(1:3))` both return false..

nalimilan: I meant isequal(rcopy(NullableArray,"c(NA,TRUE)"), NullableArray([true,true], [true,false])` and so on (i.e. convert both sides to a `NullableArray`).

PR #151, opened 2016-10-13
Fix: setup should not init '$' REPL if running with basic REPL

fedxa:
This prevents activation of REPL '$' symbol if julia is being run in a dumb terminal.  This happens if it is run e.g. within emacs (for example. usng julia-mode and ESS)

### DIFF
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -194,7 +194,7 @@ function __init__()
         # R gui eventloop
         isinteractive() && rgui_init()
         # R REPL mode
-        isdefined(Base, :active_repl) && isinteractive() && repl_init(Base.active_repl)
+        isdefined(Base, :active_repl) && isinteractive() && typeof(Base.active_repl) != Base.REPL.BasicREPL && repl_init(Base.active_repl)
     end
 
     # # IJulia hooks


src/setup.jl
```diff
@@ -194,7 +194,7 @@ function __init__()
         # R gui eventloop
         isinteractive() && rgui_init()
         # R REPL mode
-        isdefined(Base, :active_repl) && isinteractive() && repl_init(Base.active_repl)
+        isdefined(Base, :active_repl) && isinteractive() && (typeof(Base.active_repl)!=Base.REPL.BasicREPL) && repl_init(Base.active_repl)
```
ararslan: Pedantic stylistic comment: Please remove the parentheses from around the expression (it parses as desired without them) and surround `!=` with spaces.

fedxa: Done.  (Though after using & with .== recently I tend to be overprotective...)

randy3k:
Could you rebase the changes?

src/setup.jl
ararslan: Thanks! Still missing the spaces around `!=` though.

fedxa:
Ok, sorry for a bit of chaos.  I guess now all the spaces are in place, and there is one clean commit on top of master.

randy3k:
Thanks

Issue #143, opened 2016-09-01
InitError: cglobal: could not find symbol R_NilValue in library 


ScarecrowFu:
I was addingRCall

got this:
```

julia> Pkg.add("RCall")
WARNING: Pkg only works with git versions greater than v1.7.3
INFO: Initializing package repository /root/.julia/v0.4
INFO: Cloning METADATA from git://github.com/JuliaLang/METADATA.jl
INFO: Cloning cache of BinDeps from git://github.com/JuliaLang/BinDeps.jl.git
INFO: Cloning cache of Compat from git://github.com/JuliaLang/Compat.jl.git
INFO: Cloning cache of DataArrays from git://github.com/JuliaStats/DataArrays.jl.git
INFO: Cloning cache of DataFrames from git://github.com/JuliaStats/DataFrames.jl.git
INFO: Cloning cache of DataStructures from git://github.com/JuliaLang/DataStructures.jl.git
INFO: Cloning cache of GZip from git://github.com/JuliaIO/GZip.jl.git
INFO: Cloning cache of RCall from git://github.com/JuliaStats/RCall.jl.git
INFO: Cloning cache of Reexport from git://github.com/simonster/Reexport.jl.git
INFO: Cloning cache of Rmath from git://github.com/JuliaStats/Rmath.jl.git
INFO: Cloning cache of SHA from git://github.com/staticfloat/SHA.jl.git
INFO: Cloning cache of SortingAlgorithms from git://github.com/JuliaLang/SortingAlgorithms.jl.git
INFO: Cloning cache of StatsBase from git://github.com/JuliaStats/StatsBase.jl.git
INFO: Cloning cache of StatsFuns from git://github.com/JuliaStats/StatsFuns.jl.git
INFO: Cloning cache of URIParser from git://github.com/JuliaWeb/URIParser.jl.git
INFO: Installing BinDeps v0.4.4
INFO: Installing Compat v0.8.8
INFO: Installing DataArrays v0.3.8
INFO: Installing DataFrames v0.7.8
INFO: Installing DataStructures v0.4.5
INFO: Installing GZip v0.2.20
INFO: Installing RCall v0.5.1
INFO: Installing Reexport v0.0.3
INFO: Installing Rmath v0.1.2
INFO: Installing SHA v0.2.1
INFO: Installing SortingAlgorithms v0.1.0
INFO: Installing StatsBase v0.9.0
INFO: Installing StatsFuns v0.3.0
[...]

Issue #74, opened 2015-10-22
RCall initiation: could not find symbol R_BlankScalarString


egoecho:
When I run RCall in Julia 0.4.0, it have such problem. I tried Pkg.build("RCall") but it not solved. Previously RCall worked fine but the problem came after I updated anaconda (I have no idea why anaconda will lead this problem, maybe it is just coincident?).

julia> using RCall
ERROR: LoadError: InitError: cglobal: could not find symbol R_BlankScalarString in li
brary /usr/lib64/R/lib/libR.so  
 in **init** at /home/jxiong/.julia/v0.4/RCall/src/RCall.jl:71
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 in require at ./loading.jl:243
during initialization of module RCall
while loading /home/jxiong/.julia/v0.4/RCall/src/RCall.jl, in expression starting on 
line 53

simonbyrne:
This is caused by an older version of R than is currently supported, currently we require 3.2.0 or greater. What operating system are you using?

I'll try to improve the error message.

egoecho:
My R version is 3.1.0 (2014-04-10), will it be compatible?

simonbyrne:
At the moment, no. Can you try commenting out [this line](https://github.com/JuliaStats/RCall.jl/blob/711da57e7e9cc3ce3c30dec8a1f521157876d9c6/src/RCall.jl#L71) and see if it works?

What OS are you using?

egoecho:
I tried to reinstall the newest version of R (3.2.2 (2015-08-14)). When I using RCall, it crashed:

julia> using RCall

 *** caught segfault ***
address 0x54d8, cause 'memory not mapped'

Traceback:
 1: initMethodDispatch(where)
 2: fun(libname, pkgname)
 3: doTryCatch(return(expr), name, parentenv, handler)
 4: tryCatchOne(expr, names, parentenv, handlers[[1L]])
 5: tryCatchList(expr, classes, parentenv, handlers)
 6: tryCatch(fun(libname, pkgname), error = identity)
 7: runHook(".onLoad", env, package.lib, package)
 8: loadNamespace(package, c(which.lib.loc, lib.loc))
 9: doTryCatch(return(expr), name, parentenv, handler)
10: tryCatchOne(expr, names, parentenv, handlers[[1L]])
11: tryCatchList(expr, classes, parentenv, handlers)
[...]

PR #153, opened 2016-11-02
Check R library loads when searching.

simonbyrne:
Gives a better error message for #143.

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -28,8 +28,6 @@ The following will update R on recent versions of Ubuntu:
     sudo apt-get update -y
     sudo apt-get install -y r-base r-base-dev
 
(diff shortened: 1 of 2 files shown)


simonbyrne:
Should now also give a better message if the version is too old (#74).

simonbyrne:
Thoughts @randy3k ?

randy3k:
It looks good. 

PS: it seems that Julia has changed the behaviour of `isnull(Nullable(1, true))`?

simonbyrne:
Ah yes, I remember seeing that somewhere. It's now the opposite (the fields is now `hasvalue` instead of `isnull`). I can look at it tomorrow if you don't have time to do it today.

Issue #157, opened 2016-11-22
Override conversion of length-1 vectors to scalars?


dmbates:
I have run into a problem converting an `R` factor with only 1 level to a `CategoricalArrays.CategoricalArray`.

I do realize that having only one level for a factor makes it somewhat uninformative but these things do occur after subsetting a data frame and dropping unused levels.

Currently at line 30 of `src/convert-data.jl` the value of `levels` will be a scalar if there is only one level.  I can add code to check on whether the result is an array and, if not, make it an array but I was wondering if there is a more general idiom to override the conversion of length-1 vectors to scalars?

randy3k:
We could do `rcopy(Array,getattrib(s,Const.LevelsSymbol))`, it also forces the return as an array.

randy3k:
fixed, 53fb8557655bd5613db1b4bbddf5324b934a81b6

PR #158, opened 2016-11-22
Force levels of a factor to be an Array

dmbates:
See #157


randy3k:
Aternatively, we can consider `rcopy(Array, getattrib(x,RCall.Const.LevelsSymbol))`.

randy3k:
It is addressed already.

PR #159, opened 2016-12-05
WIP: change method used by getclass

simonbyrne:
Don't merge this yet: it's unclear if this function is actually meant to be used or not.

See https://stat.ethz.ch/pipermail/r-devel/2013-April/066499.html


simonbyrne:
This seems to work, though it isn't officially part of the C API. However it is exported as it is used as part of the [methods package](https://stat.ethz.ch/R-manual/R-devel/library/methods/html/methods-package.html) which is included with R.

randy3k:
It is merged.

Issue #144, opened 2016-09-04
Use NamedArrays.jl to preserve array names


nalimilan:
`rcopy(Titanic)` currently returns a plain `DataArray`, and loses all names. It would be much better to wrap this array inside a `NamedArray`.

randy3k:
It is an issue needed to be addressed. In the meanwhile, users could consider the following code.

```julia
using RCall
using NamedArrays

function RCall.rcopy(::Type{NamedArray}, r::RObject)
    dnames = rcall(:dimnames, r)
    isnull(dnames) && error("r has no dimnames")
    NamedArray(rcopy(Array, r), [rcopy(n) for n in dnames], rcopy(getnames(dnames)))
end

rcopy(NamedArray, R"Titanic")
```

phaverty:
Or AxisArrays?  Has anyone done the bake-off for those two?

Issue #138, opened 2016-08-18
support for passing custom types?


mkborregaard:
Hi,
RCall offers a great many standard conversions - julia arrays to R arrays, R lists to julia Dicts etc. But there are also numerous more complex types defined within R and julia packages. Could RCall support an API for implicitly converting between R and julia objects by allowing users to specify a conversion function?

An example: both languages have (several) representations for defining phylogenies: BioJulia has e.g. the Phylo submodule, and the ape R package has the phylo type. It is possible to use RCall code to split the julia Phylogeny type into components (arrays and such), pass them to R, then construct an ape phylo object. It is thus easy to define a conversion function.

The ideal method would be to be able to define such a method using an RCall standard API, putting them either in the relevant julia package, in RCall or an RCall_typeextensions package, and then after defining this method being able to use the conversion implicitly, e.g. 

```
R"plot($phy)"
```

simonbyrne:
That's a good idea. We should be able to do this using dispatch somehow.

simonbyrne:
I should say that though it's not documented, this is already possible in the Julia -> R direction by overloading `RCall.sexp`: just define a new method that takes your Julia type and gives the appropriate `Ptr` to an R object back. Unfortunately this requires keeping track of R garbage collection references, so perhaps isn't the best way in the long term, but I'd be happy to help out if you want to have a go at doing it.

The other direction is more complicated. At the moment we only dispatch on the R core "SEXPTYPE"s, but we could change this to dispatch on the class attribute. We would also have to figure out how to incorporate S4 classes and Reference Classes (though these seem to be less common in the wild).

mkborregaard:
I think that might be worth giving a go, but I don't think I am up to the task of dealing with R's garbage collector references. 
One thing to keep in mind is how to avoid that all packages that implement some useful type will have to depend on Rcall. One way of doing this might be for RCall to define methods that take an appropriate julia dict with all the subobjects of an R type, and then builds the R type in R. Then all packages that want to implement this functionality just need to define a function to create the specified julia dict.

simonbyrne:
One solution would be an "RCallRecipes" approach, being a lightweight repo which gets included in your package as a dependency, but doesn't require loading or installing RCall.

mkborregaard:
I like that idea a lot.

nalimilan:
That, together with a system similar to FileIO.jl, would be very useful to automatically choose what type to use when getting a special class from R.

[...]

PR #163, opened 2017-01-11
Convert Date and DateTime to/from R

lelmiger:
Support `Data` and `DateTime` types by converting from/to R's `Date` and `POSIXct` classes.

```
julia> using RCall

julia> d = Dates.today()
2017-01-11

julia> tm = now()
2017-01-11T23:14:01.148

julia> R"""
       cat(paste0("Today's date is ", $d ,".\n"))
       cat(paste0("The current timestamp is ", $tm ,".\n"))
       """;
Today's date is 2017-01-11.
The current timestamp is 2017-01-11 23:14:01.

julia> R"""
       d <- as.Date(c("2010-10-10", "2009-09-09"))
       tm <- as.POSIXct(c("1942-04-02 04:42:42", "2020-02-20 20:02:20"), "UTC")
       """;

julia> rcopy("d")
2-element Array{Date,1}:
 2010-10-10
 2009-09-09

julia> rcopy("tm")
2-element Array{DateTime,1}:
 1942-04-02T04:42:42
 2020-02-20T20:02:20
```

### DIFF
--- a/src/convert-base.jl
+++ b/src/convert-base.jl
@@ -140,6 +140,35 @@ for (J,S) in ((:Integer,:IntSxp),
     end
 end
 
+# Date and DateTime
+function sexp(d::Date)
+    res = sexp(RealSxp, convert(Float64, d - Dates.Day(719163)))
+    setclass!(res, sexp("Date"))
+    res
+end
+function sexp(a::AbstractArray{Date})
+    res = sexp(RealSxp, convert(AbstractArray{Float64}, a - Dates.Day(719163)))
+    setclass!(res, sexp("Date"))
+    res
+end
+function sexp(d::DateTime)
+    res = sexp(RealSxp, convert(Float64, d - Dates.Day(719163)) / 1000)
+    setclass!(res, sexp(["POSIXct", "POSIXt"]))
+    setattrib!(res, "tzone", sexp("UTC"))
+    res
+end
+function sexp(a::AbstractArray{DateTime})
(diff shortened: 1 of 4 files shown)


lelmiger:
Notice that conversion to/from R's `POSIXct` always occurs in UTC since julia's `DateTime` does not support time zones. This can lead to surprising results when converting `POSIXct` objects with differing or not explicitly set time zone:
```
julia> R"""tm <- as.POSIXct("2000-01-01 05:00", "Asia/Tokyo")""";

julia> rcopy("tm")
1999-12-31T20:00:00
```

nalimilan:
Cool!

Though; how about creating [`ZonedDateTime`](https://github.com/JuliaTime/TimeZones.jl) objects instead, so that time zone information is preserved?

lelmiger:
Converting `POSIXct` to `ZonedDateTime` is certainly possible.
I am not sure tough whether an additional dependency to [TimeZones.jl](https://github.com/JuliaTime/TimeZones.jl) is desirable for `RCall`.

nalimilan:
I would say yes, since that's a small, pure-Julia dependency. Anyway, we'll also need e.g. to depend on NamedArrays/AxisArrays to preserve array names (https://github.com/JuliaInterop/RCall.jl/issues/144). At some point, optional dependencies should allow requiring only the features you actually use.

randy3k:
I kind of like it, but it may complicate the work of passing custom R class object presented in https://github.com/JuliaInterop/RCall.jl/issues/138.

Issue #167, opened 2017-02-22
There is no method sexp for `SubDataFrame`


diegozea:
`MethodError: no method matching sexp(::DataFrames.SubDataFrame{Array{Int64,1}})`

I got this trying to use a R function (`lm()`) inside `DataFrames.by()`.

Best,

dmbates:
Are you planning to create a pull request?  My guess is that it may be as simple as changing a `DataFrame` signature to `Union{DataFrame,SubDataFrame}` but I haven't looked at it.

diegozea:
I changed the signature to ` sexp(d::AbstractDataFrame)` but I got: `ERROR: type SubDataFrame has no field colindex`. Is there a reason to use `d.colindex.names` insteda of `names(d)`?

randy3k:
It should be handled by the latest commits.

PR #168, opened 2017-02-22
Define sexp for AbstractDataFrame

diegozea:
It solves https://github.com/JuliaInterop/RCall.jl/issues/167

Best


randy3k:
@diegozea 
Thanks, are you able to provide some test cases too?

diegozea:
I'm not sure if https://github.com/JuliaInterop/RCall.jl/pull/169 is a good test for this, but it's the way in which I found the `sexp` error.

diegozea:
@randy3k Thanks :)

PR #173, opened 2017-03-15
v0.6 compatibility

randy3k:
I have made several changes to the 0.6 branch, the code is now compatible with both v0.5 and v0.6.
The REPL mode on v0.6 is still broken, but a [patch][1] is submitted to julia.

It will be merged into master once the tests pass.

[1]: https://github.com/JuliaLang/julia/pull/21027

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -2,9 +2,9 @@
 RCall.jl's type `Sxp` mirrors the R symbolic expression record `SEXPREC` in R API.
 These are represented by a pointer `SxpPtr` (which is called `SEXP` in R API).
 """
-abstract Sxp # SEXPREC
-typealias SxpPtr{S<:Sxp} Ptr{S} # SEXP
-typealias SxpPtrInfo UInt32 # sxpinfo_struct
+abstract type Sxp end # SEXPREC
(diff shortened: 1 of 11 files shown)


src/types.jl
```diff
@@ -253,22 +252,53 @@ RObject{RealSxp}
...
-type RObject{S<:Sxp}
-    p::Ptr{S}
-    # used for pre-defined constants
-    function RObject{S}() where S
-        new(C_NULL)
+
+# Compat.jl not yet provide support for the new syntax for inner consturctor
+# https://github.com/JuliaLang/Compat.jl/issues/332
+# and the new syntax cannot be parse in julie v0.5, therefore using string.
+if VERSION < v"0.6.0-"
```
tkelman: use a more specific cutoff for this. run contrib/commit-name.sh at the merge commit for the type system rewrite pr

nalimilan: Actually the best solution is to use the call overloading syntax, which works on both 0.5 and 0.6. See https://github.com/JuliaLang/Compat.jl/issues/332.

randy3k:
It was fixed on the master branch.

PR #175, opened 2017-03-17
major refactoring of conversion related code

### LINKED ISSUE #86: Issue related to subsetting a VectorListSxp, particularly, a StrSxp
@simonbyrne 
I notice that when I subset a StrSxp object with multiple indexes, it returns a julia array of CharSxp's. I think most users (at least myself) is expecting a StrSxp object with the corresponding indexes.
Is there any reasons for this implementation?

```julia
julia> a = RObject(["a", "b", "c"])
RCall.RObject{RCall.StrSxp}
[1] "a" "b" "c"


julia> a[1:2]
2-element Array{Any,1}:
 RCall.RObject{RCall.CharSxp}
<CHARSXP: "a">

 RCall.RObject{RCall.CharSxp}
<CHARSXP: "b">
```

Also, when only one element is being extracted from the StrSxp, the element of type CharSxp is returned to the user 
```julia
julia> a[1]
RCall.RObject{RCall.CharSxp}
<CHARSXP: "a">
```
To mirror the R behavior, should it be wrapped inside a StrSxp object and then returned to the user? This may be less important.

### LINKED ISSUE #112: Sanitize names when copying a data.frame?
I find myself needing to use

```
@rget foo
names!(foo, [Symbol(replace(string(nm), '.', '_')) for nm in names(foo)]
```

on many of the R data frames that I load.  Would it be worthwhile adding an optional argument somewhere along the line to rewrite the names?

### LINKED ISSUE #144: Use NamedArrays.jl to preserve array names

### LINKED ISSUE #174: DataFrames and DataTables support
In https://github.com/JuliaInterop/RCall.jl/pull/145 RCall was changed to create `NullableArray` and `CategoricalArray` objects rather than `DataArray` and `PooledDataArray`. This was consistent with the DataFrames plans at the time.

This decision probably needs revisiting now that the plans have changed (i.e. DataFrames will keep using `DataArray` and `PooledDataArray`, and the new DataTables package uses  `NullableArray` and `CategoricalArray`). The current state is inconsistent since `DataFrame` is not made to work well with `NullableArray` and `CategoricalArray`. So there are two options:
- keep creating `DataFrame` objects by default, and go back to creating `DataArray` and `PooledDataArray` columns
- switch to `DataTable` and keep creating `NullableArray` and `CategoricalArray` columns

I'm really torn about which solution is better. We don't have all the high-level query APIs yet to make working with DataTables as nice as it should be. Yet, going back to `DataArray` could be annoying for users who already had to adapt their code. Maybe we should do it nevertheless until the `Nullable` API stabilizes, to avoid having our data frame/table ecosystem too inconsistent or painful to use for several months.

### LINKED ISSUE #176: Stack Overflow at .../convert-base.jl:56
> ERROR: StackOverflowError:
>  in sexp(::Type{RCall.CharSxp}, ::SubString{String}) at .../convert-base.jl:56 (repeats 80000 times)

I am trying to convert a 279-element Array{SubString{String},1}: with each element containing a string. 

@rput and the R REPL give the overflow error message.

randy3k:
- [x] switch to DataArrays instead of NullableArrays (close #174)
- [x]  ~~support DataTable~~
   DataTables doesn't play well with DataFrames https://github.com/JuliaData/DataTables.jl/issues/33
    Though DataTables support is ready to merge, it was removed from the current PR at 50b1e80
 and will be merged in the future.
- [x] support AxisArrays (close #144)
- [x] support Data and time (close #163)
- [x] ~issue subsetting a StrSxp #86~ (not sure what the best practice is, won't fix now)
- [x] sanitize names in conversions (close #112)
- [x] fix #176

towards #138


nalimilan:
> support DataTable (with default being DataArrays)

If support for `DataTable` is added, it should use NullableArrays instead.

randy3k:
I meant support `DataFrame(R"mtcars")` and DataTable(R"mtcars")`. Of course, the col of DataFrame are DataArrays and that of a DataTable are NullableArrays.

nalimilan:
Then I don't understand why you wrote that. :-)

randy3k:
because I made a stupid typo...

randy3k:
Will merge soon if there are no other comments.

cc: @simonbyrne  @dmbates

REQUIRE
```diff
@@ -1,7 +1,8 @@
 julia 0.5
-DataStructures 0.4.3
+DataStructures 0.5.0
 DataFrames 0.9
-NullableArrays 0.0.10
-CategoricalArrays 0.0.6
+NullableArrays 0.1.0
+CategoricalArrays 0.1.0
+AxisArrays 0.0.6
```
nalimilan: So you've decided to use AxisArrays rather than NamedArrays? I'm still unsure what's the best approach (e.g. for my package FreqTables). The generality of `AxisArray` (numeric and categorical axes) is appealing, but the goal of the package does not seem to completely match the R approach to array names. In particular, the printing of names before the data (rather than on each line as in R/NamedArrays) isn't practical for categorical axes.

@mbauman What do you think about this? Are the goals compatible? Could the printing evolve towards what NamedArrays does? i.e. something like this:
```julia
4×4 Named Array{Int64,2}
Dim1 ╲ Dim2 │  A   B   C   D
────────────┼───────────────
a           │ 30  20  30  20
b           │ 30  20  30  20
c           │ 20  30  20  30
d           │ 20  30  20  30
```

randy3k:
Since the conversion to AxisArrays has to be explicit, i.e. `rcopy(AxisArray, foo)`. It is equally possible to support `NamedArrays` or `AxisArrays` or both.

REQUIRE
mbauman: I'd absolutely like to see better printing of AxisArrays. It's still a young project in terms of the number of hours that have been dedicated to it.

I'm not familiar with any homogenous named array in R, so I'm not entirely sure about the use-case, but the goal of AxisArrays is to provide core support for many disparate domains. Then secondary packages can layer on more domain-specific behaviors.

nalimilan: R named arrays don't have anything really specific. Names are always strings, indexing by name is supported (but only by putting the index in the position corresponding to the dimension). Names are propagated by some operations like element-wise ops and reductions (which isn't supported by AxisArrays yet IIRC). The way they are printed is similar to the `NamedArray` example above. I was afraid the way `AxisArrays` are printed was a design decision, but if you would be OK with something more like `NamedArray` it would be great.

(Then I'm not a fan of a name, I wouldn't have considered dimension names as axes, but I guess that's a minor point. :-)

PR #188, opened 2017-04-27
Fixes error when importing vector of invariant factor to Julia

kaskarn:
e.g. rcopy(R"factor(rep(1,10))"). Error due to rcopy(Array,
getattrib(s,Const.LevelsSymbol)) returning string type when importing
factor vector of repeated single level. the string is passed to
PooledDataArray constructor, which expects an array and throws error.
Fixed by enforcing Array type using rcopy(Array, getattribs(...))


randy3k:
Can you also add some tests for this issue?

kaskarn:
Added a couple tests

randy3k:
Thanks

PR #191, opened 2017-05-30
add 0.6 release and improve R installation on travis for macOS

randy3k:


randy3k:
don't have any idea why windows 32-bit build fails, cannot reproduce on my side.

PR #192, opened 2017-06-13
Implement of the RClass idea

### LINKED ISSUE #138: support for passing custom types?

randy3k:
close #138 

An example of usage can be found in [datatime.jl][1].

@richardreeve, I guess you may be interested in it.

[1]: https://github.com/JuliaInterop/RCall.jl/blob/948e1fd1d95304a78ad01150ce08389b56ed6da0/src/convert/datetime.jl

### DIFF
--- a/src/convert/base.jl
+++ b/src/convert/base.jl
@@ -1,3 +1,5 @@
+immutable RClass{Symbol} end
+
 # conversion to Base Julia types
 
 rcopy{T}(::Type{T},r::RObject; kwargs...) = rcopy(T, r.p; kwargs...)
@@ -22,6 +24,11 @@ function rcopy{T<:Number, R<:Number}(::Type{T}, x::R)
         return T(x)
     end
 end
+
+# Fallbacks
+# convert Ptr{S} to Any would use the default conversions to allow
+# automatic conversion of VecSxp objects, e.g., convert(Array{Any}, R"list(a=1, b=2)")
+rcopy{S<:Sxp}(::Type{Any}, s::Ptr{S}) = rcopy(s)
 
 # NilSxp
 rcopy{T}(::Type{T}, ::Ptr{NilSxp}) = nothing
--- a/src/convert/datetime.jl
+++ b/src/convert/datetime.jl
@@ -1,11 +1,39 @@
 # Date and DateTime
 
+# R -> Julia
+
 rcopy(::Type{Date}, s::Ptr{RealSxp}) = rcopy(Date, s[1])
 rcopy(::Type{DateTime}, s::Ptr{RealSxp}) = rcopy(DateTime, s[1])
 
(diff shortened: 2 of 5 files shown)


richardreeve:
Great, I have a brutalist version of my code (richardreeve/Phylo.jl#3) that uses this and it seems to work using `@rget`, though I'm having to create `RObject`s again in a definitely-wrong sort of way. Are there plans for exporting `RClass` and `rcall_p`, which I currently seem to need, or should I just use call `using` on them specifically?

randy3k:
`RClass` and `rcall_p` are not meant to be used by normal users. I guess we should keep them unexported.

richardreeve:
Would it be crazy to create a submodule like `RCall.Extend` for things like `protect`, `unprotect`, `rcall_p`, `RClass`, `isObject`, `isS4` to make it easy to extend the conversion functionality without having loads of `using` statements?

randy3k:
There are too many functions falling into this category. I am not sure if it becomes necessary if most of the internal functions are put in say, `RCall.API`

richardreeve:
Yes, I assumed there would be lots of them tbh, which was why I thought something like `RCall.API` would be useful. Obviously there's also a second much smaller group of functions that need `import`ing too - `rcopy`, `rcopytype`, `sexp`...

richardreeve:
So, I'm completely happy with this, and I don't want to open a can of worms, but when multiple people start providing converters for the same R type, is there going to be a way forward? As things stand (if I understand Julia multiple dispatch correctly, which I probably don't!) are they going to overwrite one another, with the last man standing? That may be fine of course, but I thought I ought to raise it...

randy3k:
I believe only the last overload counts (untested). I guess there is no easy fix for this. Since there should be one and only one output, it should not be very bad to use the last known type.

richardreeve:
Yes, that's what I thought. I guess it's okay - it would be a strange situation using two different Julia types in a single piece of code that match the same R type (and wanting to do conversions!).

Issue #206, opened 2017-10-20
Add c:/users/user/Documents/R/win-library/3.x to .libPaths


xiaodaigh:
I am running Windows 10 64bit Pro and my libraries install to "c:/Users/username/Documents/R/win-library/3.x" instead of "C:/Users/username/R/win-library/3.4" 

which was the default for me hence it would be worthwhile adding support for that

simonbyrne:
I'm not sure I understand, I don't think we change `.libPaths`. Does `.libPaths()` give you different values if run through R and RCall.jl?

xiaodaigh:
@simonbyrne I am perplexded a bit. Actually I tested it both RCall and R gives me the same .libPaths() but Rstudio gives me a different one.

So I think it's an issue with RStudio.

xiaodaigh:
Yeah. RCall.jl gives a different .libPaths()

I tried checking the R.home() and they both gave me the "same" location. Actually RCall gave me
"C:/Program Files/R/R-3.4.2"

But R and Rstudio gave me "C:/PROGRA~1/R/R-34~1.2"

simonbyrne:
Ah that helps. It looks like we're not setting HOME correctly.

simonbyrne:
Should now be fixed by #207

PR #207, opened 2017-10-23
Correctly set HOME/R_USER env vars in Windows

simonbyrne:
attempt #206 fix


simonbyrne:
I have no idea why this is failing on 0.5.

Issue #208, opened 2017-10-25
Convert R integer to Julia Int instead of Cint by default?


Non-Contradiction:
In RCall, R integer converts to Cint (Int32) by default currently. As Int32 is so little used in pure Julia code, many Julia functions on integers only accept Int (which stands for Int64 on 64 bit systems, am I correct?) as arguments. If I want to use RCall as well as other Julia packages, I need to do things like `Int()` or `Array{Int}` a lot, otherwise I will get MethodError.

Is it okay to convert R integer to Julia Int instead of Cint by default in RCall? There is going to be some performance loss, but it will save users from a lot of trouble in using `Int()` whenever the user wants to apply Julia functions to the integer part of RCall's result.

phaverty:
Going straight to Int64 on the Julia side sounds like the right thing to me.

nalimilan:
But that would mean a copy is required instead of reusing the same memory...

simonbyrne:
I think we generally copy anyway.

Non-Contradiction:
The pull request is just about this problem (?). I just change two lines in `rcopytype(s::Ptr{IntSxp})` in convert/default.jl and adjust tests accordingly.

randy3k:
Some benchmarks for those who are skeptical. 

```julia
julia> R"x = sample(1:10, 1e8, replace=T)";

julia> @time y = rcopy(Array{Int32}, R"x");
  0.261747 seconds (36 allocations: 381.471 MiB, 22.07% gc time)

julia> @time y = rcopy(Array{Int}, R"x");
  0.403379 seconds (36 allocations: 762.941 MiB, 13.22% gc time)
```

Copying to `Int` is indeed slower, but it is more user friendly.

PR #209, opened 2017-10-25
Convert R Int to Julia Int by default

Non-Contradiction:
Related to #208


randy3k:
LGTM.

src/convert/default.jl
```diff
@@ -43,9 +43,9 @@ function rcopytype(s::Ptr{IntSxp})
     elseif anyna(s)
         DataArray{Float64}
```
randy3k: opps, this should be `DataArray{Int}`. Could you also change this while you are working on it?

randy3k:
may be easier for me to change it.

PR #216, opened 2017-11-16
Add upper bounds on CategoricalArrays 0.2.0

nalimilan:
That version broke compatibility with previous releases.

See https://github.com/JuliaLang/METADATA.jl/pull/12040.

We'll have to make a PR soon to port to CategoricalArrays 0.3.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,7 +3,7 @@ Nulls 0.1.0
 DataStructures 0.5.0
 DataFrames 0.10
 DataArrays 0.5.0
-CategoricalArrays 0.1.0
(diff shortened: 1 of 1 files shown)


randy3k:
RCall actually works with CategoricalArrays 1.0 and 2.0 https://github.com/JuliaInterop/RCall.jl/blob/master/src/convert/categorical.jl#L3

Though I am not sure about 3.0.

nalimilan:
Ah, great! Then to support 0.3.0 you just need to use Missings.jl instead of Nulls.jl, and `Missing` instead of `Null`. I think you can drop support for the 0.2 series if you want. Sorry for the churn...

Issue #224, opened 2018-02-11
Unable to use R and RCall installed via conda.jl 


abhi18av:
Hi **JuliaInterop** team :)

I've been gearing up to introduce Julia to my organization where R , Python and Conda are heavily used. So, I was evaluating this package as a possible solution to tie the goals together.

I've setup conda, conda-forge and python properly 

```julia

using Conda

Conda.list()

Conda.add_channel("conda-forge")

```

Here's the issue I'm running into 

```julia 
Conda.add("r-base")

# TODO Fix the path issues
ENV["LD_LIBRARY_PATH"] = "/Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/Conda/deps/usr/lib/R/lib/libR.dylib"
Pkg.build("RCall")

using RCall
R"rnorm(10)"

```

However calling Python2 from Julia works fine 

```julia 
ENV["PYTHON"] = "/Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/Conda/deps/usr/bin/python2.7"
Pkg.build("PyCall")


using PyCall


@pyimport sys
sys.version

```

Could you please help me out to set the path properly for R ?

abhi18av:
The error message is as follows 

```
validate_libR(::String) at /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/RCall/src/setup.jl:15
 [2] locate_Rhome_libR() at /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/RCall/src/setup.jl:114
 [3] include_from_node1(::String) at ./loading.jl:576
 [4] include(::String) at ./sysimg.jl:14
 [5] include_from_node1(::String) at ./loading.jl:576
 [6] include(::String) at ./sysimg.jl:14
 [7] anonymous at ./<missing>:2
while loading /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/RCall/src/setup.jl, in expression starting on line 121
while loading /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/RCall/src/RCall.jl, in expression starting on line 41
```

randy3k:
[...]

PR #228, opened 2018-02-23
use Pkg.build to discover libR

randy3k:
see https://github.com/JuliaInterop/RCall.jl/issues/224#issuecomment-368080160

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,0 +1,54 @@
+using Compat
+
+function locate_Rhome()
+    Rhome = if haskey(ENV,"R_HOME")
+        ENV["R_HOME"]
+    else
+        try
+            readchomp(`R RHOME`)
+        catch er
+            ""
+        end
+    end
+    @static if Compat.Sys.iswindows()
+        if Rhome == ""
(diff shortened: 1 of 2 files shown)


deps/build.jl
```diff
@@ -0,0 +1,54 @@
...
+    else
+        info(abspath(filename), " has not changed")
+    end
+end
+
+Rhome = locate_Rhome()
+libR = locate_libR(Rhome)
+
+writeifchanged("deps.jl", """
+const Rhome = "$(escape_string(Rhome))"
```
simonbyrne: Maybe add a comment that this file is automatically generated?

simonbyrne:
Yeah, I was trying to be clever by trying to do without the `deps` file, but I realised this was probably a bad idea.

deps/build.jl
randy3k: done.

stevengj:
I think I have a slightly more comprehensive patch to do this...  will submit shortly.

randy3k:
Looking forward to it.

PR #229, opened 2018-02-23
move libR detection to Pkg.build, support Conda installation

### LINKED ISSUE #73: Automatically install R when required
It would be useful if we could automatically install R when no local installation is found. Here are some thoughts from having spent a morning trying and failing to get this to work:
- The R packages on linux distros are woefully out of date. The recommended way to do this is to use the CRAN repositories (e.g. see .travis.yml file). BinDeps [can't do this](https://github.com/JuliaLang/BinDeps.jl/issues/118) and probably shouldn't anyway.
- Installing the CRAN package on OS X requires administrator privileges
  - We could use the homebrew/science tap?
- CRAN doesn't provide a permalink to the latest downloads (see .travis.tml/appveyor.yml for hack workarounds)

The most promising option seems to be to use [Conda.jl](https://github.com/Luthaf/Conda.jl) (which is what PyCall.jl uses). However:
- Windows builds aren't up-to-date.
- the library (`R.dll`/`libR.dylib`/`libR.so`) is not located in a findable directory.

Getting this to work is probably going to require changing the [conda recipe](https://github.com/conda/conda-recipes/tree/master/r-base).

stevengj:
This is an alternative to #228.  It is a somewhat more comprehensive change that:

* Moves libR detection to the `Pkg.build` stage.
* Auto-installs R via Conda if it is not found.
* Remembers the libR location on subsequent builds (unless you set the `R_HOME` environment variable to override) so that you only have to configure it correctly once.

One question I had was which Conda channel to use.  There is [r-base in the r channel](https://anaconda.org/r/r-base) and [r-base in the conda-forge channel](https://anaconda.org/conda-forge/r-base).   Right now I'm using ~~conda-forge~~ `r`, but I'm not sure if there is a preference.

close #73

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,4 +6,5 @@ DataArrays 0.7.0
 CategoricalArrays 0.3.0
 AxisArrays 0.0.6
 Compat 0.34.0
+Conda
 @windows WinReg 0.2.0
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -7,40 +7,48 @@ end
 include("setup.jl")
 
 const depfile = joinpath(dirname(@__FILE__), "deps.jl")
-if isfile(depfile)
-    @eval module DepFile; include($depfile); end
-else
-    @eval module DepFile; Rhome=libR=""; end
-end
 
-if !haskey(ENV,"R_HOME") && isdir(DepFile.Rhome) && validate_libR(DepFile.libR, false)
-    Rhome, libR = DepFile.Rhome, DepFile.libR
-    info("Using previously configured R at $Rhome with libR in $libR.")
-else
-    Rhome = get(ENV, "R_HOME", "")
-    if isempty(Rhome)
-        try Rhome = readchomp(`R RHOME`); end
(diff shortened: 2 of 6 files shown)


randy3k:
I believe that `r` channel is more commonly used, though both of them are outdated. My experience is that both channels are not frequently updated, I usually end up with installing manually from CRAN.

stevengj:
I changed it to use the `r` channel.  It's not so out of date — they have R 3.4.1, only one patch behind the latest release.

randy3k:
Do we want an option to set `R_HOME` as conda R even if `R` can be located via  `PATH` variable? Maybe an environmental variable like `RCALL_USE_CONDA`?

stevengj:
As is already explained in the documentation of this PR, you can force the use of Conda by setting R_HOME to “*” (or any other non-empty invalid path).

stevengj:
Should be okay to merge?

randy3k:
Thank you. It is merged now.

Issue #238, opened 2018-04-01
cryptic warning: "incremental compilation may be broken for this module"


cecileane:
My apologies in advance for not providing a good diagnostic. I get many warnings during the precompilation of [PhyloPlots](https://github.com/cecileane/PhyloPlots.jl), which uses RCall:

```
julia> using PhyloPlots
INFO: Recompiling stale cache file /Users/ane/.julia/lib/v0.6/PhyloPlots.ji for module PhyloPlots.
WARNING: eval from module Const to PhyloPlots:    
Expr(:block, Expr(:line, 73, Symbol("/Users/ane/.julia/v0.6/RCall/src/Const.jl"))::Any, Expr(:., :NaString, Expr(:quote, :p)::Any)::Any = Expr(:call, :unsafe_load, Expr(:call, :cglobal, "R_NaString", Expr(:call, :typeof, Expr(:., :NaString, Expr(:quote, :p)::Any)::Any)::Any)::Any)::Any)::Any
  ** incremental compilation may be broken for this module **

[...]

WARNING: eval from module Const to PhyloPlots:    
Expr(:block, Expr(:line, 73, Symbol("/Users/ane/.julia/v0.6/RCall/src/Const.jl"))::Any, Expr(:., :UnboundValue, Expr(:quote, :p)::Any)::Any = Expr(:call, :unsafe_load, Expr(:call, :cglobal, "R_UnboundValue", Expr(:call, :typeof, Expr(:., :UnboundValue, Expr(:quote, :p)::Any)::Any)::Any)::Any)::Any)::Any
  ** incremental compilation may be broken for this module **
```
There is one such warning for each symbol in the loop in the definition of `load` in [Const.jl](https://github.com/JuliaInterop/RCall.jl/blob/master/src/Const.jl#L57). I checked out the master branch of PhyloPlots and PhyloNetworks, and RCall v0.10.3. Just warnings though, no error so far.

randy3k:
How about RCall v0.10.2?

cecileane:
Things look good with RCall v0.10.2, indeed: I do *not* get these warnings and all the PhyloPlots tests pass.

randy3k:
It seems like we cannot use `@eval` in `__init__()`.

randy3k:
Should be fixed by 9e3b0f1cfec357936fb5eff61c06de2e6b3aa89a.

Thanks for the report. I will make a release soon.

randy3k:
[...]

PR #242, opened 2018-04-25
update formula copying to StatsModels v0.2.4

kleinschmidt:
The representation of Formulas has changed in StatsModels.  It now holds onto the whole formula expression (e.g., `Expr(:call, :~, ...)`) as well as the parsed expression and the left- and right-hand sub-expressions.

For copying from R to Julia, it's necessary to invoke the `@formula` model to trigger parsing of the formula expression. (That's not strictly true in this release, since it's just transforming the Expr, but it will be in the future to support things like arbitrary functions in formulae).

For Julia to R, in order to match the old behavior, I propose just copying the Expr of the original (un-parsed) formula (in the `ex_orig` field).  There are some situations (where the Formula is constructed manually from left and right-hand sub expressions) that the `ex_orig` field is an empty expression, but that's deprecated now.  In those cases, this PR pulls the parsed expression in the `ex` field.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,6 +6,6 @@ DataArrays 0.7.0
 CategoricalArrays 0.3.0
 AxisArrays 0.0.6
 Compat 0.34.0
-StatsModels 0.2.0
+StatsModels 0.2.4
 Conda
 @windows WinReg 0.2.0
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -7,7 +7,7 @@ using DataArrays
 using CategoricalArrays
 using DataFrames
 using AxisArrays
-import StatsModels:Formula
+import StatsModels: Formula, @formula
 import DataStructures: OrderedDict
 
 using Compat.Dates
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -7,7 +7,7 @@ using DataArrays
 using CategoricalArrays
 using DataFrames
 using AxisArrays
-import StatsModels: Formula, @formula
+import StatsModels: Formula, parse!
 import DataStructures: OrderedDict
 
 using Compat.Dates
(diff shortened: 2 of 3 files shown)


nalimilan:
Would there be any way to do this without calling `@eval`? It sounds too bad to have to call it just to create a formula: `@eval` is allowed to run any expression with all sort of possible effects, but here we just want to get one object as a result (so it's more like a function). It sounds more like `Meta.parse` AFAICT.

randy3k:
It seems to me that the your implementation is quite fragile and is subject to be broken when StatModles removes the ex_orig field.

kleinschmidt:
Currently all `@formula` does is something like `Meta.parse`, but my vision for the future involves doing more drastic transformations like creating anonymous functions to support arbitrary transformations like `log(a)` in the formula.  AFAIK that can't be done in a function because you run into world age issues.

@randy3k there are no plans to remove the `ex_orig` field; it'll always be useful for a formula to hold onto the original expression.  And I don't see how any other solution that relies on particular internal structure of the `Formula` struct is going to be less fragile.  The other alternative would be to add getter methods for the various parts, which I'd be open to.

randy3k:
Thanks for the clarification about future plans. Going back to `@formula`, would you use `Formula(:(), :($lhs ~ $rhs), lhs, rhs) ` instead? I recently learnt that running `@eval` can have strong side effects.#238. Though it is likely that issue is not irrelevant to this PR, i still want to avoid `@eval` in function as much as possible.

kleinschmidt:
I see. We'd have to trigger parsing, too (since the `*` transformations etc. are handled at macro time in v0.2.4), but I'm okay with that for now.  It may need to change in the future if/when the parsing step becomes more involved but good to keep it incremental I think.

randy3k:
For CI fails because https://github.com/JuliaLang/METADATA.jl/pull/14404 has not been merged.

kleinschmidt:
Right, the formula tests pass locally for me with this PR.

randy3k:
I'd merge it once @dmbates adds suppport of the new version of StatModels at https://github.com/dmbates/MixedModels.jl

PR #257, opened 2018-08-13
Use new Appveyor.jl script

simonbyrne:
This is an easily updatable script

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,18 +1,18 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7-latest-win32.exe"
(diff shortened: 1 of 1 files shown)


randy3k:
How about?
```
matrix:
  allow_failures:
	- julia_version: 1
	- julia_version: nightly
```

simonbyrne:
Good point.

Actually, what are the 1.0 blockers? (sorry I haven't looked at this package for a long time)

randy3k:
`AxisArrays.jl`, see https://github.com/JuliaLang/METADATA.jl/pull/16594#issuecomment-412503646 for details

randy3k:
Given the latest update rate of AxisArrays.jl, it may still take some time for them to publish a 1.0 compatible release. What's your opinion of dropping the support of AxisArrarys (temporarily) and tag a release?

simonbyrne:
I think that seems like the best idea.

I think such "optional" dependencies should be handled with [Requires.jl](https://github.com/MikeInnes/Requires.jl)

PR #265, opened 2018-08-22
Axisarrays

randy3k:


simonbyrne:
Personally I kind of like that AxisArrays are only loaded when used: we could do the same for DataFrames and CategoricalArrays.

randy3k:
Even if we are using Requires.jl, we still need to keep the dependencies in the file [REQUIRE](https://github.com/JuliaInterop/RCall.jl/blob/master/REQUIRE) to avoid warnings complaining loading a package without requiring it.

So what are the advantages of using Requires.jl then? Users still have to download the dependencies. Given the precompiling speed of julia v1.0, I am not sure if there are any differences.

simonbyrne:
No, you shouldn't need it in `REQUIRE`, though you may need it in `test/REQUIRE` if using it during testing.

simonbyrne:
AxisArrays would still need to be added as a dependency of the project which uses it, but it doesn't need to be a dependency of RCall.

randy3k:
It is weird. I am no longer reproduce the warning.

randy3k:
So close it now and we may move all dependencies to Requires.jl.

PR #267, opened 2018-08-27
Don't use jl_value_ptr

simonbyrne:
Wrap immutable objects in a Ref to avoid the jl_value_ptr problems


randy3k:
Thank

simonbyrne:
I'm not 100% sure how necessary this is, but probably better to be on the safe side.

yuyichao:
Looks correct. A better way, if the performance of the callback is important, would be to do sth similar to https://github.com/stevengj/Cubature.jl/pull/23 to allow specialization of the callback.

PR #275, opened 2018-10-09
Sanitize symbols imported from R packages

### LINKED ISSUE #273: Cannot use functions with `.` (dot) in their names loaded using @rlibrary
MWE:
```
julia> @rlibrary PerformanceAnalytics
┌ Warning: RCall.jl:
│ Package PerformanceAnalytics (1.5.2) loaded.
│ Copyright (c) 2004-2018 Peter Carl and Brian G. Peterson, GPL-2 | GPL-3
│ https://github.com/braverock/PerformanceAnalytics
│ 
└ @ RCall ~/.julia/packages/RCall/Q4n8R/src/io.jl:110

julia> chart
chart.ACF                       chart.RollingCorrelation
chart.ACFplus                   chart.RollingMean
chart.Bar                       chart.RollingPerformance
chart.BarVaR                    chart.RollingQuantileRegression
chart.Boxplot                   chart.RollingRegression
chart.CaptureRatios             chart.Scatter
chart.Correlation               chart.SnailTrail
chart.CumReturns                chart.StackedBar
chart.Drawdown                  chart.TimeSeries
chart.ECDF                      chart.VaRSensitivity
chart.Events                    charts.Bar
chart.Histogram                 charts.BarVaR
chart.QQPlot                    charts.PerformanceSummary
chart.Regression                charts.RollingPerformance
chart.RelativePerformance       charts.RollingRegression
chart.RiskReturnScatter         charts.TimeSeries

julia> chart.Correlation
ERROR: UndefVarError: chart not defined
```

If there's no issue stopping it, I'd like to make a PR to sanitize the  symbols before being imported. Any thoughts?

jayvn:
Fixes #273 .

### DIFF
--- a/src/convert/dataframe.jl
+++ b/src/convert/dataframe.jl
@@ -1,9 +1,15 @@
 # conversion methods for DataFrames
 
-function rcopy(::Type{T}, s::Ptr{VecSxp}; sanitize::Bool=true) where T<:AbstractDataFrame
+function rcopy(::Type{T}, s::Ptr{VecSxp};
+               normalizenames::Bool=true, 
+               sanitize=nothing) where T<:AbstractDataFrame
+    if sanitize != nothing
+        Base.depwarn("The `sanitize` keyword argument is deprecated. Use `normalizenames` instead.", :rcopy)
+        normalizenames = sanitize
+    end
     isFrame(s) || error("s is not an R data frame")
     vnames = rcopy(Array{Symbol},getnames(s))
-    if sanitize
+    if normalizenames
         vnames = [Symbol(replace(string(v), '.' => '_')) for v in vnames]
     end
     DataFrame([rcopy(AbstractArray, c) for c in s], vnames)
--- a/src/convert/dataframe.jl
+++ b/src/convert/dataframe.jl
@@ -2,7 +2,7 @@
 
 function rcopy(::Type{T}, s::Ptr{VecSxp};
                normalizenames::Bool=true, 
-               sanitize=nothing) where T<:AbstractDataFrame
+               sanitize::Union{Bool,Nothing}=nothing) where T<:AbstractDataFrame
     if sanitize !== nothing
         Base.depwarn("The `sanitize` keyword argument is deprecated. Use `normalizenames` instead.", :rcopy)
         normalizenames = sanitize
--- a/src/convert/dataframe.jl
+++ b/src/convert/dataframe.jl
@@ -3,7 +3,7 @@
 function rcopy(::Type{T}, s::Ptr{VecSxp};
                normalizenames::Bool=true, 
                sanitize=nothing) where T<:AbstractDataFrame
-    if sanitize != nothing
+    if sanitize !== nothing
         Base.depwarn("The `sanitize` keyword argument is deprecated. Use `normalizenames` instead.", :rcopy)
         normalizenames = sanitize
     end
--- a/src/namespaces.jl
+++ b/src/namespaces.jl
@@ -12,7 +12,8 @@ cached_namespaces = Dict{String, Module}()
 """
 Import an R package as a julia module.
 Sanitizes the imported symbols by default, because otherwise certain symbols cannot be used.
-Eg: PerformanceAnalytics::charts.Bar in R becomes PerformanceAnalytics.charts_Bar in Julia
+
+E.g.: `PerformanceAnalytics::charts.Bar` in R becomes `PerformanceAnalytics.charts_Bar` in Julia
 ```
 gg = rimport("ggplot2")
 ```
--- a/src/namespaces.jl
+++ b/src/namespaces.jl
@@ -18,7 +18,7 @@ E.g.: `PerformanceAnalytics::charts.Bar` in R becomes `PerformanceAnalytics.char
 gg = rimport("ggplot2")
 ```
 """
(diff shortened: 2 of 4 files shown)


src/namespaces.jl
```diff
@@ -11,23 +11,32 @@ cached_namespaces = Dict{String, Module}()
 
 """
 Import an R package as a julia module.
+Sanitizes the imported symbols by default, because otherwise certain symbols cannot be used.
+Eg: PerformanceAnalytics::charts.Bar in R becomes PerformanceAnalytics.charts_Bar in Julia
```
nalimilan: Needs backticks around code. Also an empty line after the short summary would be nice, and "Eg" should be "For example" or "E.g.".

jayvn: Changed and pushed. Thanks for the review.

nalimilan:
Sorry for being a pain, but I wonder whether `sanitize` is the best name here. For example CSV.jl uses `normalizenames` for the same operation. Maybe it would make sense to use the same term for consistency?

jayvn:
> Sorry for being a pain, but I wonder whether `sanitize` is the best name here. For example CSV.jl uses `normalizenames` for the same operation. Maybe it would make sense to use the same term for consistency?
Makes sense to be consistent. But wouldn't it be better to use `sanitize` everywhere since it's shorter which in turn I assume is because it's already associated with cosmetic effects unlike `normalize` (which is why I guess they added `normalize`+`names` for clarity)? 

Anyways, no problem changing the flag. I think I'll also have to change a few  `rcopy` definitions which use `sanitize` already too.

randy3k:
I am open to rename `sanitize` to anything else. R's `read.csv`uses `check.names` fyi.

jayvn:
Hi, 
Just checking up on this PR. Should I change `sanitize` to `normalizenames` or `check.names` ?

nalimilan:
I'd be inclined to go with `normalizenames` as that's what CSV.jl uses.

randy3k:
`check.names` is definitely not a good argument name in julia btw.

jayvn:
OK. Should I change `sanitize` argument in the `rcopy` definitions as well? This'll break the existing API.

randy3k:
Yes, we should change it with a deprecation notice.

jayvn:
Sorry for the delay, having a bit of trouble with deprecating the function.

jayvn:
I cannot `@deprecate` just for a keyword argument actually. I can write a redirecting function that issues a warning if there isn't a better way.

simonbyrne:
One way to deprecate keyword args is to do something like
```julia
function rcopy(X; normalizenames::Bool=false, sanitize=nothing)
    if sanitize !== nothing
        Base.depwarn("The `sanitize` keyword argument is deprecated; use `normalizenames` instead", :rcopy)
        normalizenames = sanitize
    end
    ...
end
```

jayvn:
I just pushed the exact same snippet @simonbyrne (except that I use `!=` instead of `!==`). @fredrikekre had suggested it to me on slack yesterday. Thanks anyway.

src/convert/dataframe.jl
```diff
@@ -1,9 +1,15 @@
 # conversion methods for DataFrames
 
-function rcopy(::Type{T}, s::Ptr{VecSxp}; sanitize::Bool=true) where T<:AbstractDataFrame
+function rcopy(::Type{T}, s::Ptr{VecSxp};
+               normalizenames::Bool=true, 
+               sanitize=nothing) where T<:AbstractDataFrame
+    if sanitize != nothing
```
simonbyrne: Best to use `!==` (the union splitting optimisation doesn't work quite as well with `!=`)

jayvn: Changed and pushed.

jayvn:
Hi,
Would like to see if there's any remaining issue?

src/convert/dataframe.jl
```diff
@@ -1,9 +1,15 @@
 # conversion methods for DataFrames
 
-function rcopy(::Type{T}, s::Ptr{VecSxp}; sanitize::Bool=true) where T<:AbstractDataFrame
+function rcopy(::Type{T}, s::Ptr{VecSxp};
+               normalizenames::Bool=true, 
+               sanitize=nothing) where T<:AbstractDataFrame
```
nalimilan:
```suggestion
               sanitize::Union{Bool,Nothing}=nothing) where T<:AbstractDataFrame
```

src/namespaces.jl
```diff
@@ -11,23 +11,33 @@ cached_namespaces = Dict{String, Module}()
...
 Import an R package as a julia module.
+Sanitizes the imported symbols by default, because otherwise certain symbols cannot be used.
+
+E.g.: `PerformanceAnalytics::charts.Bar` in R becomes `PerformanceAnalytics.charts_Bar` in Julia
 ```
 gg = rimport("ggplot2")
 ```
 """
-function rimport(pkg::String, s::Symbol=:__anonymous__)
+function rimport(pkg::String, s::Symbol=:__anonymous__; normalizenames=true)
```
nalimilan:
```suggestion
function rimport(pkg::String, s::Symbol=:__anonymous__; normalizenames::Bool=true)
```

jayvn:
Thanks for the review @nalimilan

jayvn:
Hello,
Just checking to see if this is merged. Are there anymore issues?

PR #279, opened 2018-11-09
drop support of julia 0.7

randy3k:
Afterward, I will release 0.13.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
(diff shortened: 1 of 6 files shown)


mschauer:
Having 0.7 support is convenient as beginners get told to use 0.7 to figure out why their stack overflow snippets do not work.

randy3k:
I am not too concerned about this as 0.7 users can still use previous version of RCall. Plus, the julia community is moving fast.

mschauer:
Ok!

PR #281, opened 2018-11-25
1.0 is supported but 0.7 isn't.

PallHaraldsson:


PallHaraldsson:
I'm not sure this works but it should show latest 0.13.0 working. Is it valuable to see old 0.10.5 not working in 0.7?

randy3k:
Perhaps we should also remove v0.6 badge?

PR #291, opened 2019-01-03
Conversion of R data.frame with matrix columns

Non-Contradiction:


Non-Contradiction:
The failure in tests seems to be caused by the change of `view` function in `DataFrames.jl` 0.16.0.

PR #293, opened 2019-01-17
Fix for C stack checking issues on 1.1

### LINKED ISSUE #289: Error C stack usage too close to the limit (when precompiling w Julia 1.2)
Hello, I added RCall today and:

```julia
using RCall
[ Info: Precompiling RCall [6f49c342-dc21-5d91-9882-a32aef131414]
Error: C stack usage  140727995553552 is too close to the limit
Error: C stack usage  140727995553600 is too close to the limit
Error: C stack usage  140727995553520 is too close to the limit
Fatal error: unable to initialize the JIT
```

(I tried this twice, it failed both time with a similar error)

### version info

```
Julia Version 1.2.0-DEV.17
Commit 01fe7c2698 (2018-12-14 21:29 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

**Edit**:

I tried this with a couple of versions:

* 1.0.2 works
* 1.1 fails (same error)

simonbyrne:
Fixes #289.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,7 @@
 environment:
   matrix:
-  - julia_version: 1
+  - julia_version: 1.0
+  - julia_version: 1.1
   - julia_version: nightly
 
 platform:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - osx
 julia:
   - 1.0
+  - 1.1
   - nightly
 
 matrix:
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -79,10 +79,7 @@ function initEmbeddedR()
         end
 
         argv = ["REmbeddedJulia","--silent","--no-save"]
-        i = ccall((:Rf_initEmbeddedR,libR),Cint,(Cint,Ptr{Ptr{Cchar}}),length(argv),argv)
-        if i == 0
-            error("Could not start embedded R session.")
-        end
+        ccall((:Rf_initialize_R,libR),Cint,(Cint,Ptr{Ptr{Cchar}}),length(argv),argv)
 
         rs = RStart()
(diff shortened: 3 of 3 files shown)


simonbyrne:
This seems to work interactively on Mac.

randy3k:
@Non-Contradiction , I believe this doesn’t fix the issue of JuliaCall since JuliaCall does not initialize R. We might need to put the same fix at a different location for JuliaCall.

randy3k:
One more thing, on Windows, instead of `Rf_initialize_R` we should call `R_set_command_line_arguments`. Otherwise, the [`Rstart` configuration](https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/gnuwin32/embeddedR.c#L114) will be replaced.

Non-Contradiction:
Yes, adding `unsafe_store!(cglobal((:R_CStackLimit,RCall.libR),Csize_t), typemax(Csize_t))` in the `julia_setup()` fix the problem in julia REPL in JuliaCall.

randy3k:
It seems that we also need a fix for Windows. See https://github.com/JuliaInterop/RCall.jl/issues/289#issuecomment-457859721

bjce:
same issue here with `Julia 1.1` on `macOS 10.14`. What would be the step by step fix for a non programmer like me...

randy3k:
Let's merge it to master and make a release.

randy3k:
It's released https://github.com/JuliaLang/METADATA.jl/pull/21150

Issue #297, opened 2019-02-05
Build problem: LoadError: UndefVarError: different not defined  -- missing libR.so


jeback1:
I'm pointing this out in the hopes that it might make it easier for the next person who makes the same mistake I made.

I tried to install RCall on Linux. I had built R myself but had not built the shared library.  Installing RCall failed with the less-than-obvious message 'ERROR: LoadError: UndefVarError: different not defined'.  

Thank you for creating RCall.  I appreciate your willingness to share this valuable tool.

rna1990:
I had the very same issue today while building RCall. 
Thanks in advance to the RCall team for their support.

simonbyrne:
Problem appears to be this line:
https://github.com/JuliaInterop/RCall.jl/blob/08917de58017eda73499e781dfa468236485da76/deps/build.jl#L46

haberdashPI:
I'm lost... how does this fix the error? The variable `different` is still not defined in the build file. I'm getting this error when building RCall using Conda.

PR #309, opened 2019-05-29
reinsert `different` info message

haberdashPI:
Hi there, seems like #297 didn't quite get fixed by #299. Unless I'm missing something the variable `different` remains undefined in the revised version. I ran into this problem when trying to install RCall via Conda.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -21,8 +21,9 @@ try
         Rhome = get(ENV, "R_HOME", "")
         if Rhome == "*"
             # install with Conda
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -21,6 +21,7 @@ try
         Rhome = get(ENV, "R_HOME", "")
         if Rhome == "*"
             # install with Conda
+            different = "  To use a different R installation, set the \"R_HOME\" environment variable and re-run Pkg.build(\"RCall\")."
             @info "Installing R via Conda.$different"
```
simonbyrne: since it is only used here, maybe just add it directly to the message to avoid confusion.

Issue #269, opened 2018-09-14
Error with conda install on Windows


JackDunnNZ:
If I try to force the install via conda on Windows like following:

```julia
ENV["R_HOME"] = "*"
Pkg.build("RCall")
```

I get an error, because it tries to call `isdir("*")` which seems to error on Windows

randy3k:
Please check if 72dc8a1 fixes your issue.

JackDunnNZ:
Thanks for the quick response! That moves the error one step forward to `isfile("*")` in `validate_libR`

```
┌ Error: Error building `RCall`:
│ ERROR: LoadError: IOError: stat: invalid argument (EINVAL)
│ Stacktrace:
│  [1] stat(::String) at .\stat.jl:68
│  [2] isfile at .\stat.jl:303 [inlined]
│  [3] validate_libR(::String, ::Bool) at C:\Users\Jack\.julia\dev\RCall\deps\setup.jl:12
│  [4] locate_libR(::String, ::Bool) at C:\Users\Jack\.julia\dev\RCall\deps\setup.jl:53
│  [5] top-level scope at logging.jl:317
│  [6] top-level scope at C:\Users\Jack\.julia\dev\RCall\deps\build.jl:19
│  [7] include at .\boot.jl:317 [inlined]
│  [8] include_relative(::Module, ::String) at .\loading.jl:1038
│  [9] include(::Module, ::String) at .\sysimg.jl:29
│  [10] include(::String) at .\client.jl:398
│  [11] top-level scope at none:0
│ in expression starting at C:\Users\Jack\.julia\dev\RCall\deps\build.jl:10
└ @ Pkg.Operations C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v0.7\Pkg\src\Operations.jl:1068
```

randy3k:
Try again 2815fa9

cossio:
I think I'm hitting a similar error, but on Linux. 

```
julia> ENV["R_HOME"]
"*"

julia> using Pkg

julia> Pkg.build("RCall")
  Building CodecZlib → `~/julia_depot_cossio/packages/CodecZlib/DAjXH/deps/build.log`
  Building Conda ────→ `~/julia_depot_cossio/packages/Conda/uQitS/deps/build.log`
  Building RCall ────→ `~/julia_depot_cossio/packages/RCall/RPlFw/deps/build.log`
┌ Error: Error building `RCall`: 
[...]

PR #311, opened 2019-05-29
test Conda

simonbyrne:
Keep #269 fixed.


simonbyrne:
So using `ENV["R_HOME"] = "*"` causes problems when R is actually run. I think it might be better to use a variable like `ENV["R_CONDA"] = "TRUE"` or something along those lines?

simonbyrne:
Okay, this should now be working.

The main question is what to do about Conda: do we want it to install automatically if it can't find an R installation (that's what the docs say, but it isn't actually the case)?

simonbyrne:
@randy3k any thoughts?

randy3k:
Sorry for the late reply. I am busy moving.

Most Julia users trying RCall would be regular R users already. I don't think it's necessary to install R automatically.

ViralBShah:
I think the automatic installation stuff has too many challenges to get right. Might be better to remove all the Conda stuff altogether, and expect users to provide their own R.

simonbyrne:
Yeah, I think we should look into getting rid of deps altogether.

Issue #306, opened 2019-05-24
StatsModels compatibility post-Terms 2.0: Son of Terms


kleinschmidt:
Terms 2.0 (Son of Terms) (JuliaStats/StatsModels.jl#71) changes the representation of the formula rather dramatically; much of the information that was carried in the `Terms` struct is now present in the `FormulaTerm` that's created by `@formula`.  This somewhat complicates copying those objects to and from R, since the R formula is just a quoted expression.  As long as we're okay with formulae not always being fully "round trippable" then it should be possible to define sensible behavior that converts a `FormulaTerm` into an AST and (and use `@eval @formula $rformula` to do the inverse).

cc @dmbates

dmbates:
@kleinschmidt I'm running into a bit of a problem rewriting the method `rcopy(::Type{Formula}, l::Ptr{LangSxp})`  I can create the expression `ex_orig` but I am not sure how to do the equivalent of a call to the `@formula` macro on it.  In other words, I want to, within a function, take an `Expr` that has the form of a formula and create a `FormulaTerm` from it.

If I look at the code in `StatsModels` it seems that after the checks on the first argument being `:~` and the length of the arguments being 3 then I should call something like 
```
StatsModels.terms!(StatsModels.sort_terms!(StatsModels.parse!(ex_orig)))
```
but that doesn't return the type of object I want.
```
julia> StatsModels.terms!(StatsModels.sort_terms!(StatsModels.parse!(:(y ~ 1 + x))))
:(($(Expr(:escape, :~)))(Term(:y), ($(Expr(:escape, :+)))(ConstantTerm(1), Term(:x))))

julia> @formula(y ~ 1 + x)
FormulaTerm
Response:
  y(unknown)
Predictors:
  1
  x(unknown)
```

I am perhaps making this much more complicated than it needs to be.  Can you provide some guidance?

kleinschmidt:
You'll need to `eval` the expression returned by that, and apparently if you try to do that directly it chokes (because there's a lot of `esc` sprinkled into the transformed expression so it only works in a macro I guess?).  Probably the easiest thing to do is to just use the `@formula` macro:

```
julia> ex = :(y ~ 1 + x)
:(y ~ 1 + x)

julia> @eval @formula($ex)
FormulaTerm
Response:
  y(unknown)
Predictors:
  1
  x(unknown)
```

dmbates:
@kleinschmidt Your fix works, thanks.

[...]

PR #317, opened 2019-07-01
Formula transfer updated to StatsModels v0.6.0

dmbates:
Currently some of the tests are broken because the `StatsModels.@formula` macro does not preserve the original expression passed to it.  The test could be changed to compare, e.g.
```
RObject(@formula y ~ (a + b) * c)
```
to
```
R"y ~ a + b + c + a:b + b:c"
```
which would look a little peculiar or, my preferred solution, create an `exorig` field in a `FormulaTerm` like that in a `FunctionTerm`.  @kleinschmidt prefers not to do that.  See discussion in #306

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -9,8 +9,9 @@ using REPL
 using Missings
 using CategoricalArrays
 using DataFrames
+using StatsModels
 
-import StatsModels: Formula, parse!
+#import StatsModels: FormulaTerm
 import DataStructures: OrderedDict
 
 import Base: eltype, convert, isascii,
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -11,7 +11,6 @@ using CategoricalArrays
 using DataFrames
 using StatsModels
 
-#import StatsModels: FormulaTerm
 import DataStructures: OrderedDict
 
 import Base: eltype, convert, isascii,
--- a/src/convert/formula.jl
+++ b/src/convert/formula.jl
@@ -36,11 +36,9 @@ function rcopy(::Type{Expr}, l::Ptr{LangSxp})
     f
 end
 
-function rcopy(::Type{Formula}, l::Ptr{LangSxp})
-    ex_orig = rcopy(Expr, l)
-    ex = parse!(copy(ex_orig))
-    lhs, rhs = ex.args[2:3]
-    Formula(ex_orig, ex, lhs, rhs)
+function rcopy(::Type{FormulaTerm}, l::Ptr{LangSxp})
+    expr = rcopy(Expr, l)
+    @eval StatsModels.@formula($expr)
 end
 
 
@@ -65,7 +63,7 @@ sexp_formula(n::Number) = sexp(n)
 
 
 # R formula objects
-function sexp(::Type{RClass{:formula}}, f::Formula)
+function sexp(::Type{RClass{:formula}}, f::FormulaTerm)
     s = protect(sexp_formula(f.ex_orig == :() ? f.ex : f.ex_orig))
     try
         setattrib!(s, Const.ClassSymbol, "formula")
--- a/src/convert/formula.jl
+++ b/src/convert/formula.jl
@@ -29,7 +29,7 @@ function rcopy(::Type{Expr}, l::Ptr{LangSxp})
         end
         f = Expr(:call, op, cargs...)
     end
-    # unwind these opeators
+    # unwind these operators
     if op in (:+, :*, :&) && isa(f.args[2], Expr) && f.args[2].args[1] == op
         f = Expr(:call, op, f.args[2].args[2:end]..., f.args[3])
     end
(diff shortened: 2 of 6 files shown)


kleinschmidt:
Any way we solve this (other than changing the tests) is going to require some fairly deep changes to StatsModels.

Keeping the original formula expression around is a little tricky, since the formula macro doesn't actually explicitly create a call to `FormulaTerm`, it creates a call to `~` with `Term`s as arguments.  

If we want to make the formula more round-trippable (at least until `apply_schema`), one (admittedly extreme) solution would be to completely get rid of the syntactic transformations in the macro stage, and instead just create a "wrapped" version of the AST (sort of like what @oxinabox proposed for JuliaStats/StatsModels.jl#117, and JuliaStats/StatsModels.jl#119) where the symbols for called functions have been evaled into their actual function objects and other symbols are wrapped as `Term`s.  And then handle the application of special handling of `+`, `*`, and `&` at the `apply_schema` stage.

dmbates:
bump.  It would help to have this merged and a new release made.

kleinschmidt:
What's blocking here?  I don't think it's worth waiting for changes to StatsModels.  It's just not going to be possible to have full round-trip-ability of formulae between R and Julia I'm afraid because of the way that non-special function calls are handled (e.g., you can't create a formula like `@formula(y ~ 1 + log(x+y)` at run time if `log(x)` gets converted to `(x,y) -> log(x+y)`).

dmbates:
@kleinschmidt I don't think there is anything blocking.  I think it just fell through the cracks.  I didn't want to go ahead and merge it without @simonbyrne or @randy3k taking a look at it first and I imagine they are both busy.  I'm afraid I let it fall off of my ToDo list to remind them about it.

dmbates:
@kleinschmidt I don't think anything is blocking here.  I didn't want to merge without @simonbyrne or @randy3k taking a look and they are both busy, I assume.  I know Simon started a new job in the last few months.  I'm afraid I let prodding them slip off my ToDo list and I am trying to resurrect this.

dmbates:
Sorry for saying the same thing twice.  The first comment didn't appear initially so I thought I had somehow deleted it instead of hitting "Comment".

src/RCall.jl
```diff
@@ -9,8 +9,9 @@ using REPL
 using Missings
 using CategoricalArrays
 using DataFrames
+using StatsModels
 
-import StatsModels: Formula, parse!
+#import StatsModels: FormulaTerm
```
simonbyrne: should this be commented out?

src/convert/formula.jl
```diff
@@ -29,18 +29,16 @@ function rcopy(::Type{Expr}, l::Ptr{LangSxp})
...
 end
 
-function rcopy(::Type{Formula}, l::Ptr{LangSxp})
-    ex_orig = rcopy(Expr, l)
-    ex = parse!(copy(ex_orig))
-    lhs, rhs = ex.args[2:3]
-    Formula(ex_orig, ex, lhs, rhs)
+function rcopy(::Type{FormulaTerm}, l::Ptr{LangSxp})
+    expr = rcopy(Expr, l)
+    @eval StatsModels.@formula($expr)
```
simonbyrne: We shouldn't eval inside a function. Is there another interface we can use?

kleinschmidt: I'm afraid not.  At least not if we want to support calls to functions that are not part of the formula language (like `log`).

simonbyrne: the problem is that eval is called at top-level, so this probably won't actually work.

src/RCall.jl
dmbates: Now deleted

src/convert/formula.jl
dmbates: I would be tempted to merge this PR and get a new version of RCall on the registry, even if it is likely that this will not work perfectly.  Right now you can't install RCall if you want to have recent versions of GLM, MixedModels, etc. installed.

dmbates: I think it does work to call `@eval` in this case.  At least the following simple test seems okay.
```julia
julia> rcopy(R"y ~ 1 + A + B + log(C)")
FormulaTerm
Response:
  y(unknown)
Predictors:
  1
  A(unknown)
  B(unknown)
  (C)->log(C)
```

simonbyrne: The question is if it works correctly from inside a function?

dmbates: How deeply inside a function do you need?

```
julia> using RCall

R> form <- y ~ 1 + A + B + log(C)

julia> function foo()
           rcopy(R"form")
       end
foo (generic function with 1 method)

julia> foo()
FormulaTerm
Response:
  y(unknown)
Predictors:
  1
  A(unknown)
  B(unknown)
  (C)->log(C)
```

dmbates:
Any objections to my merging this?  The CI failures are caused by the inability to install the R packages on a Mac for testing.

PR #324, opened 2019-07-30
fix dataframe deprecations to access columns

ppalmes:


src/convert/dataframe.jl
```diff
@@ -22,7 +22,7 @@ function sexp(::Type{RClass{:list}}, d::AbstractDataFrame)
     rd = protect(allocArray(VecSxp, nc))
     try
         for i in 1:nc
-            rd[i] = sexp(d[nv[i]])
+            rd[i] = sexp(d[:,[nv[i]]])
```
nalimilan: I'm not sure what's the RCall policy regarding copies, but the exact equivalent of the old syntax is this, which doesn't copy the column vector:
```suggestion
            rd[i] = sexp(d[!, [nv[i]]])
```

dmbates: I think it is okay to use `!` because the vector will be copied into storage owner by the R process during the call to `sexp`.

dmbates: I am having a lot of difficulty trying to incorporate the simple `:` to `!` change.  I suspect if is because I can't change the parent branch owned by @ppalmes.  I am going to replace this pull request by another where I will acknowledge the original source.

dmbates:
Closed in favor of #327

PR #327, opened 2019-08-17
Adapt to new selection syntax in DataFrames

dmbates:
As proposed in #324 by @ppalmes but with the suggestion from @nalimilan to avoid an unnecessary copy.


dmbates:
@nalimilan Could you take another look at this and, if it looks okay, merge it?  I fixed the problems with Travis while making these changes.  Probably squash and merge.  Debugging Travis problems leads to a lot of intermediate commits.

nalimilan:
I'm not a maintainer of RCall so I wouldn't dare merging it without comments by others, but everything looks green now!

Project.toml
```diff
@@ -20,7 +20,7 @@ WinReg = "1b915085-20d7-51cf-bf83-8f477d6f5128"
 [compat]
 AxisArrays = "≥ 0.0.6"
 CategoricalArrays = "≥ 0.3.0"
-DataFrames = "≥ 0.11.0"
+DataFrames = "0.19"
```
nalimilan: Better accept later versions, which are unlikely to break things. Otherwise, we're going to end up in dependencies incompatibilities very soon.
```suggestion
DataFrames = "≥ 0.19"
```

dmbates: I have heard conflicting recommendations.  Some feel that it is safer not to use the `"≥ ..."` form in case later versions of the dependency are incompatible.  I agree that not allowing future minor or major version changes means that there is some upkeep required in keeping up with dependencies.  Perhaps that is a motivation to "minimize thine dependencies".

nalimilan: Yeah, there's no ideal solution for now. Things will be better once we tag 1.0, as then we will avoid breaking changes for a longer period. Until then I prefer using `≥` since using a strict version requirement is really painful in my experience.

dmbates:
@randy3k and @simonbyrne Do you have any objections to my making a new release with the changes for new versions of `StatsModels` and `DataFrames` that are now in the master branch?  I have heard from several people who are anxious to get an `RCall` that will work with the `terms2.0` versions of `RCall`.

The changes in #317 and #327 are now in the master branch.  I know that @simonbyrne is not happy about the use of an `eval` inside a function but I was hoping we could make a release now and fix up any infelicities later.

randy3k:
I have been lagging behind on the current update for a while because of moving and I would second if @simonbyrne agrees.

dmbates:
Thanks @randy3k.  I hope things settle down for you soon.

simonbyrne:
Thanks @dmbates for taking care of this, and tagging a new release.

PR #329, opened 2019-08-22
use --no-restore flag when initializing R

ilia-kats:
Currently, when a .RData file is present in the working directory when RCall is initialized, the RData file is loaded into the R session. This is counterintuitive and can lead to extremely high memory consumption. This patch disables this behavior.

I have only tested it on Linux. As far as I understand, RStart.RestoreAction=0 corresponds to --no-restore, so this behavior was already the default on Windows (enforced by R_SetParams in setup.jl:97)


nalimilan:
Sounds like a good idea indeed.

PR #330, opened 2019-08-23
Move test dependencies to test/Project.toml

dmbates:
In the process I found that I needed to add more test dependencies


nalimilan:
AFAIK this is only supported on Julia 1.2, so until we drop support for 1.0 and 1.1 we need to keep the current system.

ViralBShah:
I suppose this is no longer necessary.

Issue #331, opened 2019-08-24
Treating `Symbol`s like `String`s?


greimel:
I often use `ggplot2` for plotting. I often find myself in the situation, where it doesn't work, because `RCall.jl` cannot handle `Symbol`s properly.

* `CategoricalArray{Symbol}` (which occur naturally from reshaping `stack`ing dataframes) are not converted into factors.
* `Array{Symbol}` is converted to a list of symbols (dunno what that is), which cannot be processed by `ggplot2`.

```julia
using RCall, DataFrames
@rlibrary ggplot2

df_string = DataFrame(x = randn(100), y=randn(100), z1=rand(["a"; "b"], 100), z2=rand(["darkorange"; "lightblue"], 100))
ggplot(df_string) + geom_point(aes(x=:x, y=:y, color=:z1)) # works with standard colors
ggplot(df_string) + geom_point(aes(x=:x, y=:y, color=:z2)) # works with standard colors
 
df_symbol = DataFrame(x = randn(100), y=randn(100), z1=rand([:a; :b], 100), z2=rand([:darkorange; :lightblue], 100))
ggplot(df_symbol) + geom_point(aes(x=:x, y=:y, color=:z1)) # doesn't work
ggplot(df_symbol) + geom_point(aes(x=:x, y=:y, color=:z2)) # uses colors "darkred" and "lightblue"
```
I don't understand, why R now uses the symbols as colornames, not as factor categories. Is this intended?

```julia
categorical!(df_string, :z1)
ggplot(df_string) + geom_point(aes(x=:x, y=:y, color=:z1)) # works with standard colors

categorical!(df_symbol, :z1)
R"$df_symbol" # Error - malformed factor
```

Would it be possible to convert all Symbols to strings automatically before sending them to R? Could we at least do the conversion automatically for  `CategoricalArray{Symbol}`?

nalimilan:
`stack` doesn't create `CategoricalArray{Symbol}` columns AFAICT. Do you have an example?

R has the concept of symbol (see e.g. `as.symbol`), so it sounds appropriate to convert Julia symbols to that. It's unfortunate that symbols in Julia are used in many places where a string would be used in R.

The CategoricalArray issue sounds secondary to me: better decide what to do with `Array{Symbol}` first, as it is a simpler case.

bkamins:
`stack` creates `Vector{Symbol}`.

greimel:
[...]

PR #332, opened 2019-08-24
Improve conversion from CategoricalArray to factor

greimel:
Currently, only `CategoricalArray{String}` are converted to `factor` correctly. This doesn't work on `master`, but works with this PR.

`CategoricalArray{<:Number}`:
```julia
df_symbol = DataFrame(x = randn(100), y=randn(100), z1=rand([:a; :b], 100), z2=rand([:darkorange; :lightblue], 100))

categorical!(df_symbol, :x)
R"str($df_symbol)" # before: error, after: works as expected
```

`CategoricalArray{Symbol}`:
```julia
df_symbol = DataFrame(x = randn(100), y=randn(100), z1=rand([:a; :b], 100), z2=rand([:darkorange; :lightblue], 100))
ggplot(df_symbol) + geom_point(aes(x=:x, y=:y, color=:z1)) # doesn't work
ggplot(df_symbol) + geom_point(aes(x=:x, y=:y, color=:z2)) # uses colors "darkred" and "lightblue"
categorical!(df_symbol, :z2)
ggplot(df_symbol) + geom_point(aes(x=:x, y=:y, color=:z2)) # before: error - after: uses standard colors
```

The caveat is, that now a vector of `Symbol`s behaves quite differently if it's a `CategoricalVector`. (In this example above the colors change.)

This solves half of #331.

### DIFF
--- a/src/convert/categorical.jl
+++ b/src/convert/categorical.jl
@@ -23,10 +23,10 @@ function sexp(::Type{RClass{:factor}}, v::CategoricalArray)
             end
         end
         # due to a bug of CategoricalArrays, we use index(v.pool) instead of index(v)
-        setattrib!(rv, Const.LevelsSymbol, CategoricalArrays.index(v.pool))
+        setattrib!(rv, Const.LevelsSymbol, string.(CategoricalArrays.index(v.pool)))
         setattrib!(rv, Const.ClassSymbol, "factor")
         if CategoricalArrays.isordered(v)
-            rv = rcall(:ordered, rv, CategoricalArrays.levels(v))
+            rv = rcall(:ordered, rv, string.(CategoricalArrays.levels(v)))
         end
     finally
         unprotect(1)


nalimilan:
This makes sense AFAICT since R factors only allow string levels. However, due to this line:
https://github.com/JuliaInterop/RCall.jl/blob/1bd5c47ac86428eb501c017c4f534d97f21c3669/src/convert/default.jl#L266
all `CategoricalArray` objects will be converted to factors, even if their levels are not of type `String` or `Symbol`. This might be OK, but it deserves some discussion. For types other than strings, symbols and chars, it could be better to convert them to a plain vector of the corresponding type rather than to a factor.

@alyst Do you have an opinion? IIRC you work with non-string categorical vectors.

alyst:
I would rather always convert *CategoricalArray* objects to factors.
If some *CategoricalVector{Float64}* (e.g. for a *timepoint* column in an experiment) becomes *numeric*, it'll affect the generation of the model matrix for `glm()` in an unexpected way.

nalimilan:
Yeah. I guess one could argue otherwise and say that converting to string a value may destroy information (e.g. a date), but in practice we probably don't care.

alyst:
We may introduce some additional R attribute that would store the factor levels in their original type.
Probably there are even some R packages (something from *tidyverse*?) that support non-character level types and `RCall` may use their classes (IIUC, it should still work even in the absence of the package, like *data.frame* / *tibble*).

nalimilan:
Not that I know of: [forcats](https://github.com/tidyverse/forcats/) still uses standard factors (surprisingly).

PR #342, opened 2019-10-12
added a utility function to install packages

xiaodaigh:
Currently,
```julia
R"""
install.packages(pkgs)
"""
```
is failing. So I have added a function to install packages using `run`

### DIFF
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -27,7 +27,8 @@ export RObject,
    globalEnv,
    isnull, isna, anyna,
    robject, rcopy, rparse, rprint, reval, rcall, rlang,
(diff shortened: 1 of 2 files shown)


randy3k:
Try again with the master branch? `install.packages(pkgs)` should work now on Windows.

xiaodaigh:
it doesn't. stuck at 
image

randy3k:
put this in your `.Rprofile`

```
options(repos = c(CRAN = "https://cran.rstudio.com"))
```

randy3k:
It still doesn't fix the whole issue, any prompts that pop out will essentially freeze the installation.

palday:
@randy3k Would it make sense to change this to a documentation change rather than a code change?

ViralBShah:
Just tested and works fine.

PR #350, opened 2019-12-29
Fix conversion of CategoricalArray to factor vs. ordered

nalimilan:
Ordered and unordered factors were swapped. Add tests to catch this.

Bug introduced by https://github.com/JuliaInterop/RCall.jl/pull/338.


randy3k:
The appveyor failure is due to a recent release of R 4.0. It shouldn't be a big issue, but an issue of the tests. I will look into it.

PR #352, opened 2020-01-07
Add recent distribution failure handling

laborg:
See discussion: https://discourse.julialang.org/t/glibcxx-3-4-26-not-found-in-rcall/29113/9


simonbyrne:
Thanks!

PR #370, opened 2020-05-02
upper bound dependencies

randy3k:


randy3k:
close, use #362 instead.

PR #380, opened 2020-05-06
Fix installation link in readme

evanfields:
The installation link in the readme seems broken currently; this updated link (no `.html`) seems to work.


randy3k:
thanks.

PR #383, opened 2020-05-10
Faster startup

PallHaraldsson:


PallHaraldsson:
I was testing time of first plot, not that I would seriously point to people using R/ggplot instead of Julia solutions (I was just curious if actually faster), and -O1 takes 5 sec. off:
```
time julia -O1 -e 'using RCall; @time using RDatasets; @time mtcars = dataset("datasets", "mtcars"); R"library(ggplot2); ggplot($mtcars, aes(x = WT, y=MPG)) + geom_point()"'
  1.950435 seconds (3.69 M allocations: 184.399 MiB, 11.83% gc time)
  3.071046 seconds (8.16 M allocations: 379.038 MiB, 4.06% gc time)

real	0m12,314s
user	0m12,351s
sys	0m0,840s

time julia -e 'using RCall; @time using RDatasets; @time mtcars = dataset("datasets", "mtcars"); R"library(ggplot2); ggplot($mtcars, aes(x = WT, y=MPG)) + geom_point()"'
  2.663235 seconds (3.69 M allocations: 184.398 MiB, 8.58% gc time)
  6.153683 seconds (8.16 M allocations: 379.038 MiB, 1.87% gc time)

real	0m17,347s
user	0m17,355s
sys	0m0,870s
```

Note this is not actually tested, just copied Jeff's code from Plots.jl (and this only works faster, unlike -O1 from CLI), in Julia 1.5 if I recall. I just took out precompile, as it's the default, and not in e.g. Plots.jl. It's only if you do not want it when that function is useful.

You have to make sure this applies for other code to, potentially it could slow down, but I'm not sure, and if I recall this is in, or considered for PyCall.jl too.

PallHaraldsson:
I tested with a real-word package (the only R package, I know hat has yet been wrapped).

Note, this PR doesn't lower the performance of any wrapped R package in any way. It's only if you would call a R function very often (in a loop) and that R function is very fast, counted in nanoseconds, i.e. much faster than:

The R function that is called in Alpaca.jl, **takes only 0.2 sec.** but to get there, there's a lot of startup cost, in total all (I wrapped the example there in a test.jl file) **takes 30 sec**.

**I can take 73% of the time off**, or 22 sec, or realistically 7 sec (24%), and your package is responsible for part of the 7 sec overhead:

```
$ time ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia --startup-file=no test.jl
[..]
real	0m30,316s
user	0m31,401s
sys	0m1,828s

$ time ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia -O1 --startup-file=no test.jl
[..]
real	0m22,819s
user	0m24,165s

Good to know about `--compile=min` that is responsible (not -O0) for extra speedup here but can't be done with a PR:

$ time ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia -O0 --compile=min --startup-file=no test.jl
[..]
real	0m8,176s
user	0m9,635s
sys	0m1,626s
```

PallHaraldsson:
Now that Julia 1.5 is out, you should consider merging this PR. See e.g. speedup gain for RCall (and a package using it):

https://github.com/jmboehm/Alpaca.jl/issues/6

PallHaraldsson:
I would just merge the PR. If you're worried, you or your users can run the code in the PR with the new option (I think effectively disabling it):

```
$ ~/julia-1.7.0/bin/julia --min-optlevel=2
```

PR #426, opened 2021-07-24
Use values(x) to access x.data

lbilli:
In Julia 1.7 the following warning is shown when passing dataframes to R:
```julia
┌ Warning: use values(kwargs) and keys(kwargs) instead of kwargs.data and kwargs.itr
│   caller = iterate(x::Base.Pairs{Int64, RCall.ListSxp, Base.OneTo{Int64}, Ptr{RCall.ListSxp}}) at methods.jl:232
└ @ RCall ~/.julia/packages/RCall/iMDW2/src/methods.jl:232
```
This PR implements what is suggested.


simonbyrne:
Thanks!

PR #434, opened 2021-10-23
add conversion from and to NamedTuple

tomhaber:
This PR adds conversions from/to `NamedTuple`. Seems easier to use than using `Dicts`, even if it uses the same R types underneath.

When supplying the names of the `NamedTuple`, it checks if the underlying R type matches exactly. Another approach would be to only select the entries specified (if available).


ViralBShah:
Is this something that we can merge? I just happened to be here and see that this PR seems to work well with the CI and passes the new tests.

ViralBShah:
I was hoping to re-run this with the new CI - but that will need rebasing to master.

ViralBShah:
Can you rebase it to the latest master? 

Also @simonbyrne Can we merge this?

ViralBShah:
@tomhaber If it helps, also happy to give you commit access here.

simonbyrne (approved):

simonbyrne:
Thanks, sorry for the delay. I admit I don't pay much attention to this repo now.

ViralBShah:
@tomhaber Thanks for the PR and have invited you to have write permissions on this repo.

PR #442, opened 2022-02-04
switch to GitHub Actions for CI

simonbyrne:


simonbyrne:
Many thanks to @IanButterworth for the ideas in #433.

simonbyrne:
I've allowed failures for 32-bit tests on windows: theoretically this should work, but I'm not sure what is going on.

PR #445, opened 2022-02-10
Add note on installing packages via Conda

rikhuijzer:
Thanks to @adelarue in https://github.com/RcppCore/Rcpp/issues/770#issuecomment-665140387.


simonbyrne:
Maybe this should go in a "known issues" page rather than getting started?

rikhuijzer:
> Maybe this should go in a "known issues" page rather than getting started?

Yeah makes sense. It's a bit empty still but let's hope that it invites more people to submit tips and tricks.

I've tested it locally and the menu looks like this afterwards:

image

PR #456, opened 2022-10-05
remove unused type parameter

palday:
Julia 1.8.2 warns about unused type parameters


ViralBShah:
There seems to be a problem with the R install on Windows.

ViralBShah:
Ok to merge?

ararslan (approved):

PR #459, opened 2022-10-18
In CI, use the Conda-provided R

DilumAluthge:


ViralBShah:
Failing later now with similar error.

DilumAluthge:
Can you reproduce the error on a local Windows machine?

ViralBShah:
I really don't have a Windows machine - and wouldn't even know how to set up R in the same way as the CI. :-(

DilumAluthge:
We'll likely need to recruit a Windows user to help.

They don't need to set their machine up in a particular way - because we can reproduce this error in CI when `R_HOME` is set to `*`, the user can just set the `R_HOME` environment variable to `*` when they run the tests locally. No need for them to install R on their local machine.

ViralBShah:
I might be able to try that out on an underpowered windows machine I do have access to.

DilumAluthge:
> Do you know why it has issues adding General registry? https://github.com/JuliaInterop/RCall.jl/actions/runs/3277858196/jobs/5395606774#step:8:39

That looks like a bug in the https://github.com/julia-actions/julia-buildpkg action - I'll look into it.

ViralBShah:
So, 64-bit windows is only failing on the conda specific test, which we may be able to disable (or update). The 32-bit windows is failing on finding the right location of the R shared library. If we can get win64 passing, then we can disable win32 for now, and at least get to green CI.

ViralBShah:
On windows 64-bit, I set `R_HOME` to `*` and Conda failed at some point because it couldn't find `C:\Users\viral\.julia\conda\3\Library\bin\R\bin\x64\R.dll`.

```
[ Info: Installing R via Conda.  To use a different R installation, set the "R_HOME" environment variable and re-run Pkg.build("RCall").
[ Info: Downloading miniconda installer ...
[ Info: Installing miniconda ...
[ Info: Running `conda config --add channels r --file 'C:\Users\viral\.julia\conda\3\condarc-julia.yml' --force` in root environment
[ Info: Running `conda install -y 'r-base>=3.4.0,<4'` in root environment
ERROR: LoadError: Could not find library C:\Users\viral\.julia\conda\3\Library\bin\R\bin\x64\R.dll. Make sure that R shared library exists.
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:35
 [2] validate_libR(libR::String)
   @ Main C:\Users\viral\.julia\packages\RCall\6kphM\deps\setup.jl:11
 [3] locate_libR(Rhome::String)
```

However, the correct location is `/c/Users/viral/.julia/conda/3/R/bin/x64`

simonbyrne:
Does that mean `Conda.LIBDIR == "C:\Users\viral\.julia\conda\3\Library\bin"`?

simonbyrne:
Looks like it:

https://github.com/JuliaPy/Conda.jl/blob/8f7133206f3efb6308dff5a2b09393d10e6cc122/src/Conda.jl#L57

I guess we should use
```
Rhome = joinpath(Conda.ROOTENV, "R")
```

ViralBShah:
@simonbyrne Do we necessarily need to use R <= 4?

ViralBShah:
Seems like this would need more work to figure out. Just closing it again.

Issue #458, opened 2022-10-18
Windows CI cannot find R shared library on x86


ViralBShah:
I am seeing these errors in the Windows CI in recent PRs (Example: https://github.com/JuliaInterop/RCall.jl/actions/runs/3277665393/jobs/5395214932#step:7:222)

```
    Building RCall → `D:\a\RCall.jl\RCall.jl\deps\build.log`
ERROR: LoadError: Could not find library C:\R\bin\i386\R.dll. Make sure that R shared library exists.
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:35
 [2] validate_libR(libR::String)
   @ Main D:\a\RCall.jl\RCall.jl\deps\setup.jl:11
 [3] locate_libR(Rhome::SubString{String})
   @ Main D:\a\RCall.jl\RCall.jl\deps\setup.jl:43
 [4] top-level scope
   @ D:\a\RCall.jl\RCall.jl\deps\build.jl:58
 [5] include(fname::String)
   @ Base.MainInclude .\client.jl:476
 [6] top-level scope
   @ none:5
in expression starting at D:\a\RCall.jl\RCall.jl\deps\build.jl:11
ERROR: Error building `RCall`: 
```

ViralBShah:
@DilumAluthge Might you know what is going on here? We use the R setup provided on github actions by `r-lib`.

DilumAluthge:
I'm assuming that in CI, we should let RCall install R via Conda, instead of trying to find it on the system.

ViralBShah:
Should we just try to disable the whole Conda R and see what happens?

ViralBShah:
I thought that the github actions thing would set R up the right way. So perhaps either install the `r-lib` provided action or conda - but not both?

ViralBShah:
Some discussion about the right path: https://github.com/JuliaInterop/RCall.jl/pull/459#issuecomment-1284692013

palday:
I think this is stale/resolved now, but feel free to re-open if that's wrong!

PR #462, opened 2022-10-19
Make CI green - disable x86 windows and disable PPStack test on win64

ViralBShah:
Disable conda_provided_r test - This fails on Win64, but perhaps it is ok to not test, since it is just testing threading in the R installation (not RCall itself).
Also remove x86 windows from CI (Addresses #458)

cc @DilumAluthge @randy3k. I think this may be sufficient to get CI to pass and it is ok to not test 32-bit Windows, which is anyways not a Tier 1 supported Julia platform.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
-          - '1' # automatically expands to the latest stable 1.x release of Julia
+          - '1.6'
+          - '1'
           - nightly
         os: [ubuntu-latest]
         arch: [x64]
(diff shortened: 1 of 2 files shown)


randy3k:
Sorry for not being able to update and maintain RCall for a while. The test is quite important here to make sure that there is no unbalanced protect / unprotect calls. These calls are used [everywhere](https://github.com/JuliaInterop/RCall.jl/search?q=protect) in RCall. I have used it to catch a few bugs before.

Though I am not sure why it fails on Windows.

ViralBShah:
Thanks. Good to know. What if I just temporarily disable the test on Windows so that CI passes?

ViralBShah:
The tests have not passed ever on windows after the migration to github CI, from what I can see in the history.

PR #466, opened 2022-10-24
Remove Documenter key from tagbot

aviks:


simonbyrne:
Apparently it should be `SSH_KEY`

PR #471, opened 2022-12-17
fix link for ubuntu R

szcf-weiya:
- the path is outdated, and now we need to append `-cran40`, see https://cloud.r-project.org/bin/linux/ubuntu/
- so I try to add the link to the instruction on CRAN


simonbyrne:
Thanks!

PR #472, opened 2022-12-24
Allow for dispatching on the R functions

ChrisRackauckas:
In the current setup, RCall returns a closure in Julia, which means it's hard to know in Julia if something is a function from R. This makes it really hard to workaround things that are weird in R, like is seen in https://github.com/SciML/diffeqr/issues/39. This changes it to use a callable struct instead of a closure in order to allow for dispatching.

### DIFF
--- a/src/convert/base.jl
+++ b/src/convert/base.jl
@@ -193,7 +193,7 @@ end
 struct RFunction{F}
     f::F
 end
-(::RFunction)(args...) = rcopy(rcall_p(r,args...))
+(rf::RFunction)(args...) = rcopy(rcall_p(rf.f,args...))
         
 # FunctionSxp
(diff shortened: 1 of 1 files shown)


simonbyrne (approved):
We should probably make the RObject itself callable, but that may require a bit of a redesign of the interface. This is fine as an intermediate step

src/convert/base.jl
```diff
@@ -189,15 +189,20 @@ function rcopy(::Type{A}, s::Ptr{VecSxp};
     a
 end
 
-
+# Function wrapper which allows for dispatch
+struct RFunction{F}
+    f::F
+end
+(::RFunction)(args...) = rcopy(rcall_p(r,args...))
```
simonbyrne, line 196:
```suggestion
(rf::RFunction)(args...) = rcopy(rcall_p(rf.f,args...))
```

Issue #478, opened 2023-02-11
Warning when precompiling under julia v1.9.0-beta4


dmbates:
v1.9.0-beta4 and the devel version of Julia are producing warnings of the form

```
┌ RCall [6f49c342-dc21-5d91-9882-a32aef131414]
│  WARNING: method definition for sexp at /Users/dmbates/.julia/packages/RCall/Wyd74/src/convert/base.jl:214 declares type variable T but does not use it.
└ 
```

when precompiling `RCall`.

I will create a PR to fix this.

PR #479, opened 2023-02-11
Fix issue #478

dmbates:
Trivial fix that removes a warning when precompiling RCall under julia v1.9.0-beta4 and later


dmbates:
It seems that the test failures are unrelated to the change.   I can't reproduce the test failures locally.

lbilli:
Ref #456 and #474

PR #495, opened 2023-07-20
fix datetime to character conversion

palday:
The `as.character` behavior seems to have been completely removed in R 4.3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,8 @@ jobs:
           - nightly
         R:
           - 'release'
-          - '4'
-          - '3'
+          - '4.0'
+          - '3.0'
         os: [ubuntu-latest]
         arch: [x64]
         experimental: [false]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
   pull_request:
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - R ${{ matrix.R }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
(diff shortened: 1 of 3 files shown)


palday:
🤔 The failed test might be a bug in Julia 1.9.2. I don't have a Windows machine handy to bisect.

ararslan:
I'd recommend opening an issue on the Julia repo with the contents of the CI log (as requested by the error message 😄), I'm sure someone will be able to identify what's going on.

ararslan:
> The `as.character` behavior seems to have been completely removed in R 4.3.

Admittedly I don't really know how this package works so apologies for the naive question but do your changes work across R versions including 4.3 or do they require 4.3 at a minimum (or otherwise some minimum version different than whatever it currently is)?

palday:
> > The `as.character` behavior seems to have been completely removed in R 4.3.
> 
> Admittedly I don't really know how this package works so apologies for the naive question but do your changes work across R versions including 4.3 or do they require 4.3 at a minimum (or otherwise some minimum version different than whatever it currently is)?

@ararslan They should work across R 4.x versions and maybe earlier, but I would have to go back to find which version introduced `format`. Also, this is "only" in tests, not in the actual package code.

palday:
they run on R latest so yes

musvaage:
@lionel-

Time permitting would you be amenable to clarifying these matters regarding the R versions?

https://github.com/JuliaInterop/RCall.jl/pull/495#issue-1813481832

> The `as.character` behavior seems to have been completely removed in R 4.3.

https://github.com/JuliaInterop/RCall.jl/pull/495#issuecomment-1644231955

> They should work across R 4.x versions and maybe earlier, but I would have to go back to find which version introduced `format`

palday:
@ararslan I've updated the CI to run against R 3.4 (minimum supported version from `deps/setup.jl`, 4.0 and the current release). So that should answer the question about whether this impacts the R compat bounds. :sunglasses:

ararslan (approved):

Issue #490, opened 2023-06-09
Allow usage of Preferences.jl to configure R_HOME


frankier:
From https://github.com/JuliaInterop/RCall.jl/issues/480

One thing that would be a bit more robust than the current environment variable approach would be to use Preferences.jl. I assume this would be an easier change than allowing super late binding of Rhome, which would allow for it to be set from CondaPkg automatically.

I can have a go at putting together a PR if there is support for this. We could try and fall back to the environment variable approach for backwards compatibility. Would someone be willing to look at such a PR? Perhaps @palday ?

palday:
I'm not opposed to this per se, but IIRC PythonCall and PyCall also use the `ENV` variable approach? That's not a strong argument, but more something to note.

One thing to think about is what happens when there is both a Preference and ENV set and they don't agree. That type of inconsistent state can lead to very confusing debugging for users (and developers 😉).

frankier:
Fixed in https://github.com/JuliaInterop/RCall.jl/commit/31e7859ab2757d8019b42eb18522703415ff8c85

Issue #492, opened 2023-07-04
libR.so in nonstandard location


kose-y:
In a cluster environment we use, `libR.so` is located under `$R_HOME/lib64/R/lib/`, not the standard `$R_HOME/lib/`. Is there a good way to install RCall properly in this case?

frankier:
A current workaround is to modify the `libR` value in `deps/deps.jl`. If [my recent PR](https://github.com/JuliaInterop/RCall.jl/pull/496) lands it its current form, you would also be able to do this using the Julia Preferences system.

frankier:
@kose-y Now that https://github.com/JuliaInterop/RCall.jl/commit/31e7859ab2757d8019b42eb18522703415ff8c85 is merged, does this solve your problem?

kose-y:
Thanks for your amazing work. It looks like it should solve the problem.

PR #496, opened 2023-08-04
Get Rhome and libR from Preferences.jl when provided

frankier:
See: https://github.com/JuliaInterop/RCall.jl/issues/490


palday:
@frankier I haven't forgotten you! But I also probably won't review this today, my apologies. :frowning: 

Thanks for your contribution.

ParadaCarleton:
This would definitely be super useful! I'd be very happy to see it merged.

Project.toml
```diff
@@ -11,6 +11,7 @@ DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 Missings = "e1d29d7a-bbdc-5cf2-9ac0-f12de2c33e28"
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
```
palday, line 14: needs compat entry

docs/src/installation.md
```diff
@@ -29,8 +50,6 @@ Pkg.build("RCall")
 
 When `R_HOME` is set to `"*"`, RCall.jl will automatically install R for you using [Conda](https://github.com/JuliaPy/Conda.jl).
 
-Should you experience problems with any of these methods, please [open an issue](https://github.com/JuliaStats/RCall.jl/issues/new).
```
palday, line 32: I think this blurb should stay in.

docs/src/installation.md
```diff
@@ -29,8 +50,6 @@ Pkg.build("RCall")
...
-
 ## Standard installations
 
 If you want to install R yourself, rather than relying on the automatic Conda installation, you can use one of the following options:
```
palday, line 55: This might be a good spot to call out another advantage of project-specify R via preferences -- some users might have multiple R versions installed and they can now prefer different R versions for different Julia projects.

docs/src/installation.md
frankier: I moved it to the top level title "Customizing the R installation" so it's clear it's for all methods.

frankier:
Updated with your comments. I guess the tests are false negatives? They don't seem to be to do with any of the changes and weren't there before (apart from Windows latest which was on master)?

docs/src/installation.md
```diff
@@ -29,7 +50,9 @@ Pkg.build("RCall")
 
 When `R_HOME` is set to `"*"`, RCall.jl will automatically install R for you using [Conda](https://github.com/JuliaPy/Conda.jl).
 
-Should you experience problems with any of these methods, please [open an issue](https://github.com/JuliaStats/RCall.jl/issues/new).
+!!! note "Installation time R installation is shared"
+
+    When the R installation is configured at installation time, the absolute path to the R installation is currently hard-coded into the RCall.jl package, which can be shared between projects. This may cause problems if you are using different R installations for different projects which end up using the same RCall.jl installation. In this case, please [use the Preferences system instead](#Customizing-the-R-installation-at-install-time) which keeps different copies of the compiled RCall for different R installations.
```
palday, line 56: It would be good to note explicitly again that the preferences system can be used post-hoc and doesn't require a re-build.

palday: Actually, you have to call `Pkg.build()` again to activate the preferences file if you've previously built using a global install, right?

src/RCall.jl
```diff
@@ -29,13 +30,24 @@ export RObject,
    robject, rcopy, rparse, rprint, reval, rcall, rlang,
    rimport, @rimport, @rlibrary, @rput, @rget, @var_str, @R_str
 
-const depfile = joinpath(dirname(@__FILE__),"..","deps","deps.jl")
-if isfile(depfile)
-    include(depfile)
+@static if @has_preference("Rhome") || @has_preference("libR")
```
palday, line 33: ah this `@static` block means you need to do another `precompile` if you go from global R to preferences-R. But if the RCall source hasn't changed, then this won't work. Likewise, changing the preferences-R path won't be reflected. It's probably better to handle some of this logic as part of `__init__`: it will increase load-time slightly (but then again, you still have to start R afterwards, so a few milliseconds finding a path dynamically is nothing).

palday:
@frankier I'm starting to understand more of [your concerns](https://github.com/JuliaInterop/RCall.jl/pull/496#issuecomment-1677540429) a bit more. It looks like there are a few hairy details to work out still -- I think it's going to be hard to mix performance/precompilation/dynamism. I'll try to work out some more concrete suggestions than "you might have to use `__init__`", but that takes a little bit more time. 🙁 

All that said, I think this is a valuable contribution. I'm not trying to nitpick but rather help you make your work shine. We just have a little bit more work to do.

src/RCall.jl
frankier: I don't think that's needed. Have you tested this? I have a test script at https://gist.github.com/frankier/0ac468305f8b04d98c90551eda0da15d you can try. Just set RCALL_PATH to your checkout of this PR and run in from an empty directory. I see the following output:

```
...
Should be Conda R
/home/frankier/.julia/conda/3/lib/R/lib/libR.so
...
Should be CondaPkg R
/path/to/mytest/testproj/.CondaPkg/env/lib/R/lib/libR.so
...
```

The reason this works is whenever we read from a preference at compile time, it is marked as a compile time preference. I think this means that all bytecodes are then keyed off the preference value. The "global install" is keyed as `Rhome = libR = unset`. Then as soon as we set them we get a new key for our bytecode which triggers precompilation. I may have got the details wrong, but that's what I have surmised from the documentation.

frankier:
I have addressed the doc comments. For the overall approach, see my reply above.

Just to note I did take a look at trying to load stuff later on in `__init__`. This seemed like it would need quite a lot of restructuring however, since stuff from libR is used at the top level in this package so this would need quite a big restructuring. It may be marginally useful still to have a module not hardwired to a particular libR e.g. for multiple R interpreters of different versions within a particular project which could be useful if you had different independent wrapper libraries using different R code, but as far as I have seen so far it looks like this preferences approach covers the common case fairly nicely.

frankier:
@palday Have I addressed your concerns? Let me know if there's anything I can do to help with getting this merged.

palday:
@frankier Can you add another few hours to the day? 😉 😭 Mostly I just need some time to really think about this and test it out and make sure that the behavior is something we want to commit to. I like the idea, but haven't gone through all the details yet. My apologies for the delay -- it's a really nice contribution, but non trivial, which means it needs a non trivial amount of thought to review!

frankier:
Okay! Very understandable. Just checking it hadn't got lost in the mix. Cheers!

frankier:
Okay -- I've added tests for installation including switching from environment variables to preferences as an automated test. It's a bit grotty but as far as I can see this is the best way to make sure it's run as part of the test suite. Hopefully this gives some extra assurances about the approach.

schlichtanders:
@simonbyrne I am too looking for CondaPkg support for RCall - this pull request is the first needed for this. Can you please take a look and review the pullrequest?

schlichtanders:
@frankier @palday can this be merged soon?

frankier:
I can't merge this myself. I'm open to any feedback on improving the PR including adding more QA. It does have automated tests now.

schlichtanders:
A question: Does this pullrequest enable setting RHome dynamically? i.e. that R does not need to be rebuild of RHome is changed?

frankier:
This PR is designed to be as non-invasive as possible. Currently for various reasons including (IIRC) the implementation of the REPL mode, RCall.jl currently needs libR.so to compile. For this reason the Preferences system is used to create a compile-time preference. Compile time preferences give us the exact behavior we want: it will automatically retrigger a package compilation when the libR location is changed and it will keep different compiled copies for different libRs.

As it is, I would say this PR improves on the current situation, but libR still cannot be changed at runtime (I guess this is what is meant by dynamic), which might be preferable if possible. The alternative of making libR not needed at compile time might need a fairly big refactor possibly resulting in multiple packages. I don't have the resources for this myself, and I don't think the development resources seem readily forthcoming from elsewhere.

frankier:
@palday I don't want to nag, nor to overstep, but I was wondering whether it might be possible to expedite this by finding someone else with commit access to this repo willing to review + merge this or to merge it yourself following a final check from some other trusted party? (In the latter case, I can try to find someone myself.)

palday:
> @palday I don't want to nag, nor to overstep, 

Definitely not nagging and frankly your patience with my very limited attention is something I wish all contributors would have. :heart: Truly the right mix of patience and 'bump'ing!

My sincere apologies for not being a better shepherd here -- I haven't had the mental space for big FOSS work lately, especially for things that aren't part of my day job. :slightly_frowning_face: 

> but I was wondering whether it might be possible to expedite this 1) by finding someone else with commit access to this repo willing to review + merge this OR 2) to merge it yourself following a final check from some other trusted party? (In the latter case, I can try to find someone myself.)

Having a second opinion would definitely make me feel more comfortable. I can ping @simonbyrne, @randy3k and @dmbates as erst while maintainers of this package, but I don't know how much time/attention they have to spare at the moment.

src/setup.jl
```diff
@@ -171,6 +171,13 @@ end
 include(joinpath(dirname(@__FILE__),"..","deps","setup.jl"))
 
 function __init__()
+    if Rhome == ""
+        error(
+            "No R installation was detected as RCall installation time. " *
```
palday, line 176:
```suggestion
            "No R installation was detected at RCall installation time. " *
```

frankier:
Hi @palday! Thanks for asking in case anyone was around. I also asked on Slack and did not get a response. It looks like we're own our own!

I wanted to post a link to this talk, in particular "principle 9" given here: https://youtu.be/fQJrEQIRPZU?t=3547

What do you think? I have "skin in the game" in that I am using R packages quite a bit together with Julia, and am likely to continue to for quite a while since it's part of the ecosystem I have no desire to reimplement from scratch.

While I think there are no obvious problems, I understand your reservations since it's really easy to break people's workflows especially with the installation stage. What I can agree to do is to monitor this issues and try to address them, adding more tests as I go. To this end, I can also add an issue template mentioning that I can be pinged for any issues to do with installation or configuring which R installation to use.

To summarise one more time the reasons I think this PR should be merged

 1) **Makes sure that telling RCall.jl to use different R installations for different projects doesn't mean that projects accidentally clobber each other's preferences** (I think this is quite important, since it can cause pretty confusing problems for people)
 2) Allows the location of R to be specified after installation
 3) Allows having libR.so in a non-standard location see https://github.com/JuliaInterop/RCall.jl/issues/492
 4) Following on from 1 and 2, allows some kind of integration with RCall.jl, although as-per https://github.com/JuliaPy/CondaPkg.jl/issues/121 this may continue to require a manual step in some cases at least for the time being.

Even given the proviso of 4, I believe the changes made in this PR are still something we would want to integrate when making further improvements, since preferences should usually be the preferred way of configuring packages and overriding their defaults. I will probably continue to look into improving things after this PR is merged.

test/installation.jl
palday, line 1: I like testing the installation in more depth :smile:

test/installation.jl
```diff
@@ -0,0 +1,148 @@
...
+# Test whether we can install RCall without R
+rcall_without_r_proj = mk_temp_proj()
+with_subinterpreter_task("can_install_rcall_without_r") do
+  process = restart_in_proj(rcall_without_r_proj)
+  @test process.exitcode == 0
+end
+
+# We want to guard this with a version check so we don't run into the following
+# (non-widespread) issue on older versions of Julia:
+# https://github.com/JuliaLang/julia/issues/34276
```
palday, line 125:
```suggestion
# https://github.com/JuliaLang/julia/issues/34276
# (related to incompatible libstdc++ versions)
```

I used to run into this issue all the time....

test/runtests.jl
```diff
@@ -1,6 +1,8 @@
 using RCall
 using Test
 
+include("installation.jl")
```
palday, line 4:
```suggestion
@testset "installation" begin 
    include("installation.jl")
end
```

test/installation.jl
frankier: Great! Hopefully it addresses some of the possible failure modes you were bringing up before as well as making sure we can make future improvements. It should be reasonable easy to add more now the test harness is there in case you have any more ideas.

palday: I have a way to simplify this test set a fair amount -- do you mind if I push directly to your PR? It's a bit much to turn into a bunch of code suggestions.

docs/src/installation.md
```diff
@@ -8,6 +8,28 @@ Pkg.add("RCall")
 
 ## Customizing the R installation
 
+There are two ways to configure the R installation used by RCall.jl:
+
+  * [Using Julia's Preferences system](#Customizing-the-R-installation-using-Julia's-Preferences-system)
+  * [At install time, using the `R_HOME` environment variable](#Customizing-the-R-installation-at-install-time)
```
palday, line 14: nit: this is at _build_ time for RCall. You can rebuild a package that's already been installed. (IIRC TimeZones also takes advantage of this to make its data are up to date.)

docs/src/installation.md
```diff
@@ -8,6 +8,28 @@ Pkg.add("RCall")
...
+
+```toml
+[RCall]
+Rhome = "/path/to/env/lib/R"
+libR = "/path/to/env/lib/R/lib/libR.so"
+```
+
+!!! note
+    When these preferences are set, they take precedence over the R installation configured at installation time.
+
```
palday, line 30: It might be good to add a note here with the method used in the installation tests as a way to use CondaPkg-packaged R as the preferred route.

src/RCall.jl
```diff
@@ -29,13 +30,24 @@ export RObject,
...
+    @static if !(@has_preference("Rhome") && @has_preference("libR"))
+        error("RCall: Either both Rhome and libR must be set or neither of them")
```
palday, line 35: not in this PR, but I would like to see the ability to figure out well-behaved libR when only Rhome is set

src/RCall.jl
```diff
@@ -29,13 +30,24 @@ export RObject,
...
-    error("RCall not properly installed. Please run Pkg.build(\"RCall\")")
+    const depfile = joinpath(dirname(@__FILE__),"..","deps","deps.jl")
+    if isfile(depfile)
+        include(depfile)
+    else
+        error("RCall not properly installed. Please run Pkg.build(\"RCall\")")
+    end
 end
 
+__precompile__(Rhome != "")
```
palday, line 49: it's very rare to see explicit `__precompile__` directives nowadays -- what happens if you don't have this?

src/RCall.jl
palday: :thinking: It still feels weird to have a `@static` call for something that's dynamic. My preference would be drop it here because 
- it feels weird to me
- `@static` at the top-level of a module for something that's not platform dependent, doesn't actually get you that much (because the module definition is only evaluated once unless you do something fancy like explicitly invoking `Base.require` or wonky with Revise.jl)

But that's just my preference.

frankier: That's a good point! I'm not sure why I set that, except I wanted to make extra-sure things would be at compile time, but as you say there's no need and as a result it's a bit odd. I have added a comment about what is going on here and pulled them out into temporary variables to defeat short-circuiting.

docs/src/installation.md
frankier: I suppose I was trying to contrast the two from the user's perspective, so that I could use consistent shorthand on this page. In a sense talking about build-time instead would also be misleading, since in terms of being a correct description of the implementation, R_HOME and the Preferences approach are *both* configured at build-time, but Preferences has the advantage that it triggers a build at the correct time and keeps multiple builds around. I have switched to using the `R_HOME` environment variable as a shorthand.

docs/src/installation.md
frankier: Good idea. Something better should be done here at some point, but I think a "cookbook"-style note is probably best for the meantime.

src/RCall.jl
frankier: I guess having a default would be fine, but yes please, let's defer it for now.

src/RCall.jl
frankier: This is to abort precompilation when there is not R available so that it is possible to install RCall.jl without an error when there is no R available, in case the user wants to set up preferences between installation and RCall import time. The reason for aborting precompilation is actually more about the dependent packages, since when they are precompiled they will import RCall and `__init__` will run and not be able to complete. In case this behaviour is surprising, see: https://github.com/frankier/precompiletest

I've added a message to this effect and re-written the code to make it more clear that this is about aborting precompilation rather than the old "opting-in" to precompilation approach.

frankier:
Thanks for cleaning up the tests and for the feedback.

It looks like this test is failing https://github.com/JuliaInterop/RCall.jl/actions/runs/7444228626/job/20250301671?pr=496

Last time I looked it was failing due to a problem present on `main`, but now it seems to be an actual error. I guess the location of `R.dll` is being guessed incorrectly? Do you have any ideas? I guess it wasn't introduced during the test refactoring? I don't have a convenient way to test with Windows locally, but will try to set-up/fire-up a VM or something to troubleshoot in case the answer to those two questions is a negative.

palday:
> Thanks for cleaning up the tests and for the feedback.
> 
> It looks like this test is failing https://github.com/JuliaInterop/RCall.jl/actions/runs/7444228626/job/20250301671?pr=496
> 
> Last time I looked it was failing due to a problem present on `main`, but now it seems to be an actual error. I guess the location of `R.dll` is being guessed incorrectly? Do you have any ideas? I guess it wasn't introduced during the test refactoring? I don't have a convenient way to test with Windows locally, but will try to set-up/fire-up a VM or something to troubleshoot in case the answer to those two questions is a negative.

Same here -- I've got a Windows VM set up, but I'm really slow figuring it out. AFAICT the path is correct, but there's something about loading DLLs not in the system variable `%PATH%` that doesn't work. I haven't yet figured it out, but I'm working on it when I get a few minutes in the evenings. :slightly_frowning_face: 

There was also an issue with the CI on macOS, but it looked to be a more fundamental issue with Conda and the availability of `md5` vs `md5sum` vs something else, so I just disabled that check. The last bit to get this merged is understanding the limitations on different platforms (e.g. "Linux can handle all this dynamically, but Windows requires setting XXX") and having the CI reflect that. Once we have that, I think this is good to go! So close!

frankier:
Hi! I just checked this on a Windows laptop I got access to. Using dependencywalker.exe, I found that BLAS wasn't being found which led me to realise that this was actually a bug in the test code that was only surfacing on Windows. The CondaPkg environment is not activated(!) Apparently on the other systems another version of BLAS just gets picked up. On non-Conda R it's not a problem because everything is compiled into the DLLs I think. I fixed this and added a note to the docs. I also marked the information about using with CondaPkg as experimental.

frankier:
Okay -- another failure on Windows. I feel like it worked correctly locally but I will look into it soon.

palday:
> Okay -- another failure on Windows. I feel like it worked correctly locally but I will look into it soon.

I feel like we're heading in the right direction -- thanks for your patience and efforts!

docs/src/installation.md
```diff
@@ -8,7 +8,58 @@ Pkg.add("RCall")
...
+end
+set_preferences!(RCALL_UUID, "Rhome" => target_rhome, "libR" => target_libr)
+```
+
+!!! note
+    So that CondaPkg managed R finds the correct versions of its shared library dependencies, such as BLAS, you must arrange for the Conda environment to be active when `RCall` is imported so that native library loading paths are set up correctly. If you do not do so, it is still possible that things will appear to work correctly if compatible versions are available from elsewhere in your library loading path, but the resulting code can break in some environments and so is not portable.
+
+    At the moment there are two options for arranging for this:
+    1. Use `CondaPkg.activate!(ENV)` to permanently modify the environment
+    2. Use `CondaPkg.withenv()`, in which case you will have to use `Base.require(Main, :RCall)` to import RCall rather than `using`.
```
palday, line 58: you can still do `using`, but you'll need to do

```julia
CondaPkg.withenv() do
    @eval using RCall
end
```

docs/src/installation.md
```diff
@@ -29,7 +80,11 @@ Pkg.build("RCall")
 
 When `R_HOME` is set to `"*"`, RCall.jl will automatically install R for you using [Conda](https://github.com/JuliaPy/Conda.jl).
 
-Should you experience problems with any of these methods, please [open an issue](https://github.com/JuliaStats/RCall.jl/issues/new).
+When `R_HOME` is set to `"_"`, you opt out of all attempts to automatically locate R. RCall.jl will not be importable until you set a location for R [using the Preferences system](#Customizing-the-R-installation-using-Julia's-Preferences-system).
+
+!!! note "R_HOME-based R installation is shared"
+
```
palday, line 86:
```suggestion
```

docs/src/installation.md
```diff
@@ -29,7 +80,11 @@ Pkg.build("RCall")
 
 When `R_HOME` is set to `"*"`, RCall.jl will automatically install R for you using [Conda](https://github.com/JuliaPy/Conda.jl).
 
-Should you experience problems with any of these methods, please [open an issue](https://github.com/JuliaStats/RCall.jl/issues/new).
+When `R_HOME` is set to `"_"`, you opt out of all attempts to automatically locate R. RCall.jl will not be importable until you set a location for R [using the Preferences system](#Customizing-the-R-installation-using-Julia's-Preferences-system).
```
palday, line 83: this would actually work for any string that is not a path nor a `*`, right?

test/installation/swap_to_prefs_and_condapkg.jl
```diff
@@ -39,6 +39,7 @@ set_preferences!(RCALL_UUID,
                  "Rhome" => target_rhome, "libR" => locate_libR(target_rhome))
 RCall = nothing
 CondaPkg.withenv() do
+    Pkg.build("RCall")
```
frankier, line 42: Is this still needed?

src/setup.jl
```diff
@@ -171,13 +171,12 @@ end
 include(joinpath(dirname(@__FILE__),"..","deps","setup.jl"))
 
 function __init__()
-    if Rhome == ""
-        error(
+    # This should actually error much sooner, but this is just in case
+    isempty(Rhome) && error(
```
frankier, line 175: Ha! Mixed feelings about this sneaky hack to pass coverage. The reason there is no coverage here is because when we shell out in `installation.jl` coverage stops being measured. It's probably not worth tracking it.

frankier:
Thanks for the docs feedback. I just left a couple of small double-checking comments back and then I guess merge time?

test/installation/swap_to_prefs_and_condapkg.jl
palday: It seemed to be in my testing.

src/setup.jl
palday: Yeah, it's definitely a hack but it's one I'm comfortable with for such "defense in depth" type measures.

palday (approved):

PR #502, opened 2023-09-25
Fix complete_line for Julia REPL with tab-completion

frankier:
The current nightly Julia REPL asserts the type of LineEdit.complete_line


palday:
Thanks @frankier! Can you bump the patch version so that we can release this right away?

frankier:
Done

PR #515, opened 2024-01-04
Add support for tuples and better round tripping of named tuples

palday:
[JuliaCall](https://github.com/Non-Contradiction/JuliaCall/blob/c05473bea78a0197c639f7e82ab1c6f2e943e1cc/inst/julia/convert.jl#L7-L11) currently commits some type piracy to provide support for tuples:

```julia
function sexp(x ::Tuple)
    r = sexp(Array{Any}([x...]))
    setattrib!(r, :class, sexp("JuliaTuple"))
    r
end
```

This PR upstreams a more efficient (and memory safe) approach that avoids the type piracy. The type piracy is particularly unfortunate because it breaks [JellyMe4](https://github.com/palday/JellyMe4.jl/issues/72) when used from within JuliaCall. (JuliaCall works by starting Julia and running RCall from the Julia session.)

I also take advantage of the new tuple implementation to simplify the named tuple implementation a bit and potentially make it a bit more efficient.

Finally, I take advantage of R's class-via-attributes system to annotate the lists created from (named) tuples as coming from Julia, which makes it possible to default to constructing (named) tuples, i.e. the original types, when roundtripping them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ docs/build
 docs/gh-pages
 docs/site
 deps/build.log
+lcov.info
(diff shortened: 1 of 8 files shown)


ararslan (approved):
Makes sense AFAICT

PR #516, opened 2024-04-01
Fix possible typo in docs

dahong67:
Hello! I've been excited to see the progress on making it possible to use RCall with CondaPkg.

Tried taking it for a spin using the directions in the docs, and I think I've found a typo. Figured I'd make a quick PR in case it is indeed a typo.


palday (approved):
good catch, thanks!

PR #523, opened 2024-05-02
handle empty input for reval_p

schlichtanders:
if nothing is actually given (like an empty string) this now behaves correctly in returning nothing


palday (commented):
@schlichtanders can you add a test of this behavior?

schlichtanders:
@palday yes, please may you point me to the place in the code where you would like to have the test?

palday:
@schlichtanders maybe just at the top of https://github.com/JuliaInterop/RCall.jl/blob/master/test/basic.jl ?

Issue #508, opened 2023-11-18
If julia functions which was brought to R throws an error, no real error is thrown, but an error object is returned #216


schlichtanders:
I created an [issue at JuliaCall](https://github.com/Non-Contradiction/JuliaCall/issues/216) however after a first inspection, I think it is actually an RCall issue.

Concretely something in here must not raise julia errors when run on Python. Also the stacktrace is not reported.
https://github.com/JuliaInterop/RCall.jl/blob/69c22ddc91a69d3411fa282652bbe00b5d4bdea4/src/callback.jl#L122-L147

### Helpful steps to provide error support

Looking into JuliaCall's error handling, there are two places:
1. on JuliaCall julia side the following try catch converts a julia error to an R object https://github.com/Non-Contradiction/JuliaCall/blob/c05473bea78a0197c639f7e82ab1c6f2e943e1cc/inst/julia/setup.jl#L165-L194
```julia
function docall(call1)
    try
        # ...
    catch e
        Rerror(e, stacktrace(catch_backtrace())).p;
    end;
end
```

2. on JuliaCall R side the following intercepts these R objects and raises them as R errors https://github.com/Non-Contradiction/JuliaCall/blob/c05473bea78a0197c639f7e82ab1c6f2e943e1cc/R/interface.R#L76-L78

```R
    r <- .julia$do.call_(jcall)

    if (inherits(r, "error")) stop(r)
```

Both these steps would need to be replicated inside the conversion of a Julia Function to an R function.

schlichtanders:
I found where the julia error object is transformed into a simpleError, skipping stacktrace entirely

https://github.com/JuliaInterop/RCall.jl/blob/69c22ddc91a69d3411fa282652bbe00b5d4bdea4/src/callback.jl#L61

this should be changed to the RCall equivalent of JuliaCall's
```julia
err = Rerror(e, stacktrace(catch_backtrace())).p;
```

schlichtanders:
The second step needed is to surround the following with the `if (inherits(r, "error")) stop(r)` logic

https://github.com/JuliaInterop/RCall.jl/blob/69c22ddc91a69d3411fa282652bbe00b5d4bdea4/src/callback.jl#L132-L135

Unfortunately, I couldn't find a way yet how to add such if else logic from Julia.

schlichtanders:
I fixed the one part for me now, but it seems `juliaCallback` cannot be fixed outside of RCall itself. 

[...]

PR #525, opened 2024-05-02
raise julia errors

schlichtanders:
fixes https://github.com/JuliaInterop/RCall.jl/issues/508

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -100,6 +100,9 @@ const juliaCallback = RObject{ExtPtrSxp}()
 function setup_callbacks()
     julia_extptr_callback_ptr = @cfunction(julia_extptr_callback,Ptr{UnknownSxp},(Ptr{ListSxp},))
     juliaCallback.p = makeNativeSymbolRef(julia_extptr_callback_ptr)
+
+    # helper to throw Julia errors on R side
+    reval("...stop_if_error <- function (obj) if (inherits(obj, 'error')) stop(obj) else obj")
 end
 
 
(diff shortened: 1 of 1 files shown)


src/callback.jl
```diff
@@ -124,24 +127,31 @@ Wrap a callable Julia object `f` an a R `ClosSxpPtr`.
...
+                fptr,
+                Const.DotsSymbol
+            )
+        )
+    )
     nprotect = 2
     local clos
     try
-        args = protect(sexp_arglist_dots())
+        args = RCall.protect(RCall.sexp_arglist_dots())
```
palday, line 149:
```suggestion
        args = protect(sexp_arglist_dots())
```

src/callback.jl
```diff
@@ -124,24 +127,31 @@ Wrap a callable Julia object `f` an a R `ClosSxpPtr`.
...
         nprotect += 1
-        lang = rlang_p(:function, args, body)
-        clos = reval_p(lang)
+        lang = RCall.rlang_p(:function, args, body)
+        clos = RCall.reval_p(lang)
```
palday, line 152:
```suggestion
        lang = rlang_p(:function, args, body)
        clos = reval_p(lang)
```

palday (commented):
I'm generally onboard with this, but it needs tests of the new behavior. :smile:

src/callback.jl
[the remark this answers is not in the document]
schlichtanders: thank you for all the changes. yes indeed I was previously overwriting it but as this got discouraged like an error in julia 1.10 I switched to forking RCall. These were apparently some leftovers

schlichtanders:
@palday I took away the `RCall.` qualifier. Can you point me to the place where you would like to have the test?

ViralBShah:
I believe the test to have is something that checks for the new julia error being raised as expected in this PR, so we don't regress.

palday:
@schlichtanders there is an entry with the comment "callbacks" in the tests folder:

https://github.com/JuliaInterop/RCall.jl/blob/master/test/basic.jl#L67

That seems like a good place to put it. :smile:

Issue #529, opened 2024-05-31
Difficulty installing RCall on a low version of julia 


isaamael:
Hi ！
I had to install RCall in Julia 1.4.0 because of the R package dependency, I ran the following code, however I'm having trouble resolving the error

```
julia> ENV["R_HOME"] = "~/miniconda3/envs/RTIGER/lib/R"
"~/miniconda3/envs/RTIGER/lib/R"

julia> Rhome = "~/miniconda3/envs/RTIGER/lib/R"
"~/miniconda3/envs/RTIGER/lib/R"

julia> libR = "~/miniconda3/envs/RTIGER/lib/R/lib/libR.so"
"~/miniconda3/envs/RTIGER/lib/R/lib/libR.so"

julia> Pkg.build(PackageSpec(name="RCall", version="0.13.2"))
   Building Conda → `~/.julia/packages/Conda/sDjAP/deps/build.log`
   Building RCall → `~/.julia/packages/RCall/ffM0W/deps/build.log`
┌ Error: Error building `RCall`: 
│ ERROR: LoadError: R_HOME is not a directory.
│ Stacktrace:
│  [1] error(::String) at ./error.jl:33
│  [2] top-level scope at /public/home/xuruiqiang/.julia/packages/RCall/ffM0W/deps/build.jl:42
│  [3] include(::String) at ./client.jl:439
│  [4] top-level scope at none:5
│ in expression starting at /public/home/xuruiqiang/.julia/packages/RCall/ffM0W/deps/build.jl:10
└ @ Pkg.Operations /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:892
```

R is installed via conda and I have determined that "~/miniconda3/envs/RTIGER/lib/R" is the folder.
I specified the RCall version, because the newer versions of RCall don't work on Julia before 1.6.0+, I don't know if 0.13.2 will, I didn't find instructions corresponding to the julia version! I'd like to try it, but the installation comes first!

Seeking help !

palday:
Instead of using the `~`, can you use the full path? `~` is something that Bash provides and it's not necessarily recognized at the file system level. Looking at the rest of your error message, you need something like

```julia
julia> ENV["R_HOME"] = "/public/home/xuruiqiang/miniconda3/envs/RTIGER/lib/R"
```

isaamael:
Dear Developer,  @palday 
[...]

PR #530, opened 2024-06-04
bump Julia compat

palday:
There's nothing in RCall itself that's incompatible with Julia 1.0, but the dependencies for testing are not resolvable on 1.0 and we haven't tested on 1.0 in a while.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -29,7 +29,7 @@ Preferences = "1"
 Requires = "0.5.2, 1"
 StatsModels = "0.6, 0.7"
 WinReg = "0.2, 0.3, 1"
-julia = "1"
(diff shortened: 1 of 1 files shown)


ararslan (approved):

ViralBShah:
Should we tag a new release to go with this?

palday:
@ViralBShah I think the real fix to #529 would be updating compat in General to reflect when the dependencies are actually resolvable. Releasing a new version with a bumped compat bound will just lead to Pkg trying the older versions and still failing in the same way. I haven't had time to actually bisect and figure out at what point we can no longer resolve our dependencies -- I'll try to do that in the next few days.

PR #532, opened 2024-07-16
Deal with normalization collisions in reimport

### LINKED ISSUE #418: ERROR: invalid redefinition of constant with R library
Hello!

I am trying to load the r package [ape](https://cran.r-project.org/web/packages/ape/index.html) from within julia using `RCall`.

It fails when trying to load it with the `@rlibrary` macro:

```julia
julia> @rlibrary ape
ERROR: invalid redefinition of constant node_depth_edgelength
Stacktrace:
 [1] top-level scope
   @ none:1
 [2] eval
   @ ./boot.jl:360 [inlined]
 [3] rimport(pkg::String, s::Symbol; normalizenames::Bool)
   @ RCall ~/.julia/packages/RCall/3mHXJ/src/namespaces.jl:41
 [4] rimport
   @ ~/.julia/packages/RCall/3mHXJ/src/namespaces.jl:22 [inlined]
 [5] rimport (repeats 2 times)
   @ ~/.julia/packages/RCall/3mHXJ/src/namespaces.jl:46 [inlined]
 [6] top-level scope
   @ ~/.julia/packages/RCall/3mHXJ/src/namespaces.jl:95
```

however, it works when calling library inside an R string (i.e., using `R""`):

```julia
julia> R"library(ape)"
RObject{StrSxp}
[1] "ape"       "stats"     "graphics"  "grDevices" "utils"     "datasets" 
[7] "methods"   "base"     
```
My guess is that something happens when loading ape's namespace but I don't understand why the macro fails while the R string does not. I have been unable to find a similar case in earlier issues.

Thanks in advance,

Gustavo

palday:
- give more informative error messages
- allow customization of normalization
- fix a few small things while hitting that code

closes #418


src/namespaces.jl
```diff
@@ -28,7 +31,12 @@ function rimport(pkg::String, s::Symbol=:__anonymous__; normalizenames::Bool=tru
         m = Module(s, false)
         id = Expr(:const, Expr(:(=), :__package__, pkg))
         if normalizenames
-            exports = [Symbol(replace(x, '.' => '_')) for x in members]
+            exports = [Symbol(replace(x, normalization...)) for x in members]
+            if length(exports) > length(unique(exports))
+                dupes = [k for (k, v) in countmap(exports) if v > 1]
```
ararslan, line 36: You're identifying unique elements twice; you could `countmap` first and check whether any have a count > 1. Also note that `length(x) > length(unique(x))` can be written more succinctly as `!allunique(x)`.

palday: 🤦‍♂️

ararslan (approved):
`countmap` is such a simple function that I almost feel like it could be more convenient to basically inline it rather than load StatsBase (assuming it isn't already a transitive dependency, which it probably is). Or perhaps separate this chunk out into something like
```julia
function normalized_names(members, replacements...)
    occurrences = Dict{Symbol,Int}()
    for member in members
        x = Symbol(replace(member, replacements...))
        occurrences[x] = get(occurrences, x, 0) + 1
    end
    duplicated = [k for (k, v) in occurrences if v > 1]
    if !isempty(duplicated)
        throw(ArgumentError("Names are not unique after normalization: " * join(duplicated, ", "))
    end
    return collect(keys(occurrences))
end
```
But anyway, doesn't matter to me, do what feels right.

palday:
It is already a transitive dependency (StatsModels), which is why I decided to just grab it that way. But I like your minor refactor, so I'll update to go that way. 😄

palday:
HUH, so it turns out that the exports are order sensitive ?! That ordering is lost when we do do `collect(keys(...))`, so everything breaks.

PR #533, opened 2024-07-17
allow conversion (with loss of precision) for arbitrary precision POSIXct objects

### LINKED ISSUE #396: Dealing with microseconds
R permits fractional seconds in POSIXct objects, seemingly to arbitrary accuracy. But DateTime only goes to millisecond accuracy.

```
julia> using RCall

julia> rcopy(R"as.POSIXct('2020-10-09 12:09:46.123')")
2020-10-09T11:09:46.123

julia> rcopy(R"as.POSIXct('2020-10-09 12:09:46.123456789')")
ERROR: InexactError: Int64(6.373792498612346e13)
Stacktrace:
 [1] Int64 at ./float.jl:710 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] Millisecond at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Dates/src/types.jl:33 [inlined]
 [4] rcopy at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/convert/datetime.jl:9 [inlined]
 [5] rcopy(::Type{Dates.DateTime}, ::Ptr{RealSxp}) at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/convert/datetime.jl:6
 [6] rcopy(::Ptr{RealSxp}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{,Tuple{}}}) at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/convert/default.jl:14
 [7] rcopy at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/convert/default.jl:9 [inlined]
 [8] #rcopy#24 at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/convert/default.jl:6 [inlined]
 [9] rcopy(::RObject{RealSxp}) at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/convert/default.jl:6
 [10] top-level scope at /Users/sthomas/.julia/packages/RCall/AEOQ7/src/macros.jl:71
```

I work with timestamps from BigQuery that have microsecond accuracy, so I can't get the results from a query without hitting that error.

Is there a way to tell RCall to truncate to the accuracy of the DateTime type when converting? I don't actually need the microseconds, I just want the data to fit into a DateTime.

palday:
closes #396

### DIFF
--- a/docs/src/conversions.md
+++ b/docs/src/conversions.md
@@ -74,6 +74,16 @@ r = robject(d)
 rcopy(r)
 ```
 
+Note that R's `POSIXct` supports higher precision than DateTime:
+
+```@example 1
+r = reval("as.POSIXct('2020-10-09 12:09:46.1234')")
+```
+
+```@example 1
+d = rcopy(r)
+```
+
 ## DataFrames
 
 ```@example 1
--- a/docs/src/conversions.md
+++ b/docs/src/conversions.md
@@ -84,6 +84,10 @@ r = reval("as.POSIXct('2020-10-09 12:09:46.1234')")
 d = rcopy(r)
 ```
 
+!!! note "Conversions to `DateTime` are given in UTC!"
+    `POSIXct` stores times internally as UTC with a timezone attribute.
+    The conversion to `DateTime` necessarily strips away timezone information, resulting in UTC values.
+
 ## DataFrames
 
 ```@example 1
--- a/src/convert/datetime.jl
+++ b/src/convert/datetime.jl
@@ -6,8 +6,14 @@ rcopy(::Type{Date}, s::Ptr{RealSxp}) = rcopy(Date, s[1])
 rcopy(::Type{DateTime}, s::Ptr{RealSxp}) = rcopy(DateTime, s[1])
 
 rcopy(::Type{Date}, x::Float64) = Date(Dates.UTInstant(Dates.Day((isnan(x) ? 0 : x) + 719163)))
-rcopy(::Type{DateTime}, x::Float64) =
-    DateTime(Dates.UTInstant(Dates.Millisecond(((isnan(x) ? 0 : x) + 62135683200) * 1000)))
+function rcopy(::Type{DateTime}, x::Float64)
+    ms = ((isnan(x) ? 0 : x) + 62135683200) * 1_000
+    if !isinteger(ms)
+        @warn "Precision lost in conversion to DateTime"
+    end
+    ms = round(Int, ms)
+    return DateTime(Dates.UTInstant(Millisecond(ms)))
+end
 
 # implicit conversion `rcopy(d)`.
 function rcopytype(::Type{RClass{:Date}}, s::Ptr{RealSxp})
--- a/src/convert/datetime.jl
+++ b/src/convert/datetime.jl
@@ -8,11 +8,12 @@ rcopy(::Type{DateTime}, s::Ptr{RealSxp}) = rcopy(DateTime, s[1])
 rcopy(::Type{Date}, x::Float64) = Date(Dates.UTInstant(Dates.Day((isnan(x) ? 0 : x) + 719163)))
 function rcopy(::Type{DateTime}, x::Float64)
     ms = ((isnan(x) ? 0 : x) + 62135683200) * 1_000
-    if !isinteger(ms)
+    ms_int = round(Int, ms)
+    if ms != ms_int
         @debug "Precision lost in conversion to DateTime"
     end
-    ms = round(Int, ms)
-    return DateTime(Dates.UTInstant(Millisecond(ms)))
+
+    return DateTime(Dates.UTInstant(Millisecond(ms_int)))
 end
 
 # implicit conversion `rcopy(d)`.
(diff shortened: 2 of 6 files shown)


src/convert/datetime.jl
```diff
@@ -6,8 +6,14 @@ rcopy(::Type{Date}, s::Ptr{RealSxp}) = rcopy(Date, s[1])
 rcopy(::Type{DateTime}, s::Ptr{RealSxp}) = rcopy(DateTime, s[1])
 
 rcopy(::Type{Date}, x::Float64) = Date(Dates.UTInstant(Dates.Day((isnan(x) ? 0 : x) + 719163)))
-rcopy(::Type{DateTime}, x::Float64) =
-    DateTime(Dates.UTInstant(Dates.Millisecond(((isnan(x) ? 0 : x) + 62135683200) * 1000)))
+function rcopy(::Type{DateTime}, x::Float64)
+    ms = ((isnan(x) ? 0 : x) + 62135683200) * 1_000
+    if !isinteger(ms)
+        @warn "Precision lost in conversion to DateTime"
```
ararslan, line 12: Though R users are accustomed to warnings, they're rather unusual for Julia. Perhaps it would make sense for this to be either an error or a `@debug` message? Any idea what the impact would be of the precision loss in the worst case?

palday: Hmmmm.... erroring is in some sense the current behavior, but we could add a more meaningful error message that the user needs to round to millisecond precision first.

R essentially allows for infinite precision here (I couldn't find a documented precision), but Julia restricts this to milliseconds. So you're never more than 1ms off, but whether that matters depends a lot on application domain. Looking at the POSIX roots of this, POSIX mostly cares about seconds, so 🤷. 

I think `@debug` is fine. It's better rounding than what a user would get if they rounded in R, then did the conversion to Julia:
```
R> x <- as.POSIXct('2020-10-09 12:09:46.1234')

R> round(x, "millisecond")
Error in match.arg(units) : 
  'arg' should be one of “secs”, “mins”, “hours”, “days”, “months”, “years”
```

ararslan: Any idea how common it is to have `POSIXct` values in R with meaningful sub-millisecond precision? At any rate, it could be worth calling this out somewhere in the documentation.

palday: https://github.com/JuliaInterop/RCall.jl/pull/533/commits/3766f29272bccab282041152d0e6eb94f6fbfbf4

src/convert/datetime.jl
```diff
@@ -6,8 +6,14 @@ rcopy(::Type{Date}, s::Ptr{RealSxp}) = rcopy(Date, s[1])
...
 
 rcopy(::Type{Date}, x::Float64) = Date(Dates.UTInstant(Dates.Day((isnan(x) ? 0 : x) + 719163)))
-rcopy(::Type{DateTime}, x::Float64) =
-    DateTime(Dates.UTInstant(Dates.Millisecond(((isnan(x) ? 0 : x) + 62135683200) * 1000)))
+function rcopy(::Type{DateTime}, x::Float64)
+    ms = ((isnan(x) ? 0 : x) + 62135683200) * 1_000
+    if !isinteger(ms)
+        @debug "Precision lost in conversion to DateTime"
+    end
+    ms = round(Int, ms)
```
ararslan, line 14: Since you're rounding anyway, you could skip the `trunc` that `isinteger` does and just check for equality of values pre- and post-`round`. Doesn't really matter though.

test/convert/datetime.jl
```diff
@@ -135,3 +135,8 @@ r = RObject(d)
 @test ismissing(rcopy(r)[2])
 @test rcopy(Array, r)[[1,3]] == d[[1,3]]
 @test ismissing(rcopy(Array, r)[2])
+
+# microseconds on R side #396
+@test_logs((:debug, "Precision lost in conversion to DateTime"),
+            min_level=Logging.Debug,
+            rcopy(R"as.POSIXct('2020-10-09 12:09:46.1234')"))
```
ararslan, line 142: It'd probably be worthwhile to test the result as well.
```suggestion
            rcopy(R"as.POSIXct('2020-10-09 12:09:46.1234')"))
@test rcopy(R"as.POSIXct('2020-10-09 12:09:46.1234')") == DateTime("2020-10-09T12:09:46.123")
```

ararslan (approved):

docs/src/conversions.md
```diff
@@ -74,6 +74,20 @@ r = robject(d)
...
+r = reval("as.POSIXct('2020-10-09 12:09:46.1234')")
+```
+
+```@example 1
+d = rcopy(r)
+```
+
+!!! note "Conversions to `DateTime` are given in UTC!"
+    `POSIXct` stores times internally as UTC with a timezone attribute.
+    The conversion to `DateTime` necessarily strips away timezone information, resulting in UTC values.
```
ararslan, line 89: If `POSIXct` has a notion of timezone, I wonder if it should actually be represented on the Julia side as a `ZonedDateTime` rather than a `DateTime`. That'd be a breaking change but something possibly worth considering, at least if I understand the situation correctly.

src/convert/datetime.jl
palday: yep! hence #535

docs/src/conversions.md
palday: yep, hence #535!

PR #539, opened 2024-07-18
Clean up and add docstrings

palday:


src/types.jl
```diff
@@ -21,172 +46,436 @@ abstract type VectorListSxp <: VectorSxp end
...
+- `head::SxpHead`
+"""
+struct DotSxp <: Sxp
     head::SxpHead
 end
 
-"R \"any\" object"
-struct AnySxp <: Sxp     # type tag 18
+"""
+    DotSxp <: Sxp
```
dmbates, line 334: Should be `AnySxp`, right?

src/callback.jl
```diff
@@ -1,7 +1,13 @@
...
-It is probably safe to such to make the external pointer.
+    makeNativeSymbolRef(fptr::Ptr{Cvoid})
+
+Register a function pointer as an R `NativeSymbol`.
+
+!!! note
+   `R_MakeExternalPtrFn` is not part of the official R API before
+    R 3.4 and we are supposed to use `R_registerRoutines` instead.
+    It is probably safe to such to make the external pointer, even
+    on older versions of R (although we don't officially support them.)
```
ararslan, line 10: I find this a bit difficult to follow as written. Is the following a correct understanding of the situation?

- R < 3.4
    - We don't support this R version
    - The functionality of `R_MakeExternalPtrFn` is exposed officially by `R_registerRoutines`
    - `R_MakeExternalPtrFn` does exist but is not part of R's official API
    - This Julia function might work with this version of R but we don't particularly care one way or another
- R ≥ 3.4
    - We support this version of R
    - `R_MakeExternalPtrFn` is part of R's official API
    - This Julia function is expected to work and it's a bug if it doesn't

If we don't support versions of R where the R function this Julia function wraps isn't officially supported by R then is it worth mentioning at all?

src/callback.jl
```diff
@@ -14,20 +20,30 @@ function makeNativeSymbolRef(fptr::Ptr{Cvoid})
...
+end
 
 """
-The function called by R .External for Julia callbacks.
+    julia_extptr_callback(p::Ptr{ListSxp})
+
+The function called by R `.External` for Julia callbacks.
 
-It receives a `Ptr{ListSxp}` containing
+The argument should be`Ptr{ListSxp}` containing
```
dmbates, line 41: Space before first backtick?

dmbates (commented):
Thanks for cleaning this code up.  I have a couple of minor suggestions.

I noticed that the tests on nightly are failing, apparently because of a change in the REPL implementation.  I assume the failure is unrelated to these changes.  Is that the case?

dmbates:
I'll let @palday handle the merging of the changes in base into this branch.  Those kinds of git operations make me want to run away and hide.

src/callback.jl
ararslan:
```suggestion
The argument should be a `Ptr{ListSxp}` containing
```

src/types.jl
ararslan:
```suggestion
    AnySxp <: Sxp
```

src/types.jl
```diff
@@ -21,172 +46,436 @@ abstract type VectorListSxp <: VectorSxp end
...
+struct AnySxp <: Sxp
     head::SxpHead
 end
 
-"R list (i.e. Array{Any,1})"
-struct VecSxp <: VectorListSxp     # type tag 19
+"""
+    VecSxp <: VectorListSxp
+
+Represnetation of R list (i.e. Array{Any,1}).
```
ararslan, line 350:
```suggestion
Representation of R list (i.e. `Array{Any,1}`).
```

src/types.jl
```diff
@@ -241,73 +536,101 @@ eltype(::Type{StrSxp}) = Ptr{CharSxp}
...
-for (i,T) in enumerate(typs)
+Ordered collection of R `SEXP` types, so that the (index - 1)
+matches the type tag.
+"""
+const SXP_TYPES = (NilSxp, SymSxp, ListSxp, ClosSxp, EnvSxp,
+              PromSxp, LangSxp, SpecialSxp, BuiltinSxp, CharSxp,
+              LglSxp, Nothing, Nothing, IntSxp, RealSxp,
+              CplxSxp, StrSxp, DotSxp, AnySxp, VecSxp,
+              ExprSxp, BcodeSxp, ExtPtrSxp, WeakRefSxp, RawSxp,
+              S4Sxp)
```
ararslan, line 606:
```suggestion
const SXP_TYPES = (NilSxp, SymSxp, ListSxp, ClosSxp, EnvSxp,
                   PromSxp, LangSxp, SpecialSxp, BuiltinSxp, CharSxp,
                   LglSxp, Nothing, Nothing, IntSxp, RealSxp,
                   CplxSxp, StrSxp, DotSxp, AnySxp, VecSxp,
                   ExprSxp, BcodeSxp, ExtPtrSxp, WeakRefSxp, RawSxp,
                   S4Sxp)
```

src/types.jl
```diff
@@ -241,73 +536,101 @@ eltype(::Type{StrSxp}) = Ptr{CharSxp}
...
         @eval sexpnum(::Type{$T}) = $(i-1)
     end
 end
 
 
 """
-Convert a `Ptr{UnknownSxp}` to an appropriate `Ptr{S<:Sxp}`.
+    sexp(p::Ptr{UnknownSxp})
+
+Return a restrictively parameterized `Ptr{<:Sxp}` pointing the same object.
```
ararslan, line 618:
```suggestion
Return a restrictively parameterized `Ptr{<:Sxp}` pointing to the same object as `p`.
```

src/types.jl
```diff
@@ -241,73 +536,101 @@ eltype(::Type{StrSxp}) = Ptr{CharSxp}
...
-sexp(::Type{S},s::Ptr{S}) where S<:Sxp = s
-sexp(::Type{S},r::RObject{S}) where S<:Sxp = r.p
+"""
+   sexp(::Type{<:Sxp}, s::RObject{<:Sxp})
+
+Return the associated `Sxp` pointer.
+"""
+sexp(::Type{S}, r::RObject{S}) where S<:Sxp = r.p
+# do we need this method?
+# sexp(::Type{S}, s::Ptr{S}) where S<:Sxp = s
```
ararslan, line 636: Probably doesn't hurt and possibly avoids some unnecessary indirection but I can't say for certain

ararslan:
> I assume the failure is unrelated to these changes. Is that the case?

Yes, nightly has been failing on all recent PRs.

src/callback.jl
palday: correct. tbh, I had considered omitting it completely because the comment was largely historical

src/types.jl
palday: ah yes, the joy of copy-paste when doing lots of docstrings..

src/types.jl
palday: Yeah, this is why I left here but commented out. I don't think it breaks anything (and all tests pass...) but just in case we get a complaint in the next few weeks...

I'm hoping to make this less relevant in general by slowing move more to dealing with `RObject` and deprecating/removing/hiding the pointer (`Sxp`) interface.

src/callback.jl
palday, line 10:
```suggestion
```

src/types.jl
```diff
@@ -213,23 +505,26 @@ mutable struct RObject{S<:Sxp}
         r
     end
     # SymSxps are not garbage collected, so preserve not necessary.
-    RObject{S}(p::Ptr{SymSxp}) where S = new{S}(p)
+    RObject{SymSxp}(p::Ptr{SymSxp}) = new{SymSxp}(p)
```
ararslan, line 508: This isn't functionally the same as what was written but I'm not sure whether what was written before was intentional. It was implicitly relying on `convert(Ptr{S}, p::Ptr{SymSxp})` but I don't know enough about how this package works to know whether that's a sensible conversion to do. With this change, `RObject{S}(::Ptr{SymSxp})` for `S != SymSxp` will be a `MethodError`, won't it?

src/types.jl
ararslan: > and all tests pass

Worth noting that the package doesn't have 100% coverage, though I haven't looked at whether this definition was covered.

src/types.jl
palday: Yeah, I think it will be a method error, but I think that having a method error there makes sense. The usual call would be `RObject(p)`, which invokes this definition:

```julia
RObject(p::Ptr{S}) where S<:Sxp = RObject{S}(p)
```

and thus would always use the narrowest possible construction.

src/types.jl
palday:
```suggestion
# do we need this method?
sexp(::Type{S}, s::Ptr{S}) where S<:Sxp = s
```

palday: re-enable method

ararslan (approved):

Issue #444, opened 2022-02-09
Get rid of raw pointer (Ptr{XSxp}) interfaces


simonbyrne:
We are a bit cavalier about passing around raw pointers (sexp) objects, and so we don't always ensure that a reference is kept when doing `ccall`s: this is likely the cause of the occasional segfaults that are reported (e.g. #348, #431).

The easiest option is probably to change all functions to use `RObject`s directly, and define appropriate `cconvert`/`unsafe_convert` methods.

PR #540, opened 2024-07-19
safer indexing for AbstractArray

palday:
~closes # 431~

apparently not on Mac

### DIFF
--- a/src/convert/base.jl
+++ b/src/convert/base.jl
@@ -114,7 +114,7 @@ end
 function rcopy(::Type{Vector{Bool}},s::Ptr{LglSxp})
     a = Array{Bool}(undef, length(s))
     v = unsafe_vec(s)
-    for i = 1:length(a)
+    for i in eachindex(a, v)
         a[i] = v[i] != 0
     end
     a
@@ -122,7 +122,7 @@ end
 function rcopy(::Type{BitVector},s::Ptr{LglSxp})
     a = BitArray(undef, length(s))
     v = unsafe_vec(s)
-    for i = 1:length(a)
+    for i in eachindex(a, v)
         a[i] = v[i] != 0
     end
     a
@@ -135,7 +135,7 @@ end
 function rcopy(::Type{Array{Bool}},s::Ptr{LglSxp})
     a = Array{Bool}(undef, size(s)...)
     v = unsafe_vec(s)
-    for i = 1:length(a)
+    for i in eachindex(a, v)
         a[i] = v[i] != 0
     end
     a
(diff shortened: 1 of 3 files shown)


src/convert/base.jl
```diff
@@ -264,8 +264,8 @@ sexp(::Type{RClass{:character}},st::AbstractString) = sexp(RClass{:character}, s
 function sexp(::Type{RClass{:character}}, a::AbstractArray{T}) where T<:AbstractString
     ra = protect(allocArray(StrSxp, size(a)...))
     try
-        for i in 1:length(a)
-            ra[i] = a[i]
+        for (i, idx) in enumerate(eachindex(a))
+            ra[i] = a[idx]
```
ararslan, line 268: Note that this still assumes 1-based indexing for `ra` since `enumerate` will start with 1.

palday: yes, but we create `ra` and it's indeed one based

ararslan: Fair, though we also create `a` and `v` as 1-based arrays in the above `rcopy` methods which means the loop specifications there were safe prior to using `eachindex`. I assumed you were going for minimal assumptions but if not then that's fine.

ararslan:
```
MethodError: no method matching keys(::Ptr{StrSxp})
```
Looks like something is getting `eachindex`ed that shouldn't be (or is missing a suitable definition, perhaps)

src/convert/base.jl
palday: `typeof(ra)` is lacking a `eachindex` method and the fallback calls `keys`; I'm going to take a stab at defining something appopriate

src/methods.jl
```diff
@@ -99,6 +99,7 @@ function iterate(s::Ptr{S}, state) where S<:VectorSxp
     (s[state], state)
 end
 
+Base.eachindex(s::Ptr{<:VectorSxp}) = Base.OneTo(length(s))
```
ararslan, line 102: Feels weird to define `eachindex` for a `Ptr` but we also have `iterate` and `getindex`...

src/convert/base.jl
```diff
@@ -296,8 +298,10 @@ end
 function sexp(::Type{RClass{:list}}, a::AbstractArray)
     ra = protect(allocArray(VecSxp, size(a)...))
     try
-        for i in 1:length(a)
-            ra[i] = a[i]
+        # we want this to work even if a doesn't use one-based indexing
+        # we only care about ra having the same length (which it does)
+        for (i, idx) in zip(eachindex(ra), eachindex(a))
+            ra[i] = a[idx]
```
ararslan, line 304: This feels like a situation where a `copyto!` method could make sense, assuming such a method doesn't already exist (I assume it doesn't)

ararslan (approved):

src/methods.jl
palday: yeah, it does, but then again we're treating R native arrays as iterators without copying...

src/convert/base.jl
palday: Hmmm, yeah, but I'm going to punt on that until I've tackled #444

PR #542, opened 2024-07-24
IJulia docstrings + tests

### LINKED ISSUE #141: Test IJulia components
We really need a good way to do this. Any ideas?

palday:
closes #141

### DIFF
--- a/test/ijulia.jl
+++ b/test/ijulia.jl
@@ -4,9 +4,12 @@ using RCall
 using Weave
 using Test
 
-Conda.add("jupyter")
-testpath = Base.Fix1(joinpath, @__DIR__)
 jupyter_path = joinpath(Conda.BINDIR, "jupyter")
+
+if !isfile(jupyter_path)
+    Conda.add("jupyter")
+end
+testpath = Base.Fix1(joinpath, @__DIR__)
 Weave.notebook(testpath("ijulia.jmd"); out_path=@__DIR__, jupyter_path=jupyter_path)
 
 run(`$(jupyter_path) nbconvert $(testpath("ijulia.ipynb")) --to html --embed-images`)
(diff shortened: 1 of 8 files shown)


test/ijulia.jl
```diff
@@ -0,0 +1,40 @@
...
+
+# create a folder ijulia_files with the exported images -- could be useful if we ever set up percy
+# run(`$(jupyter_path) nbconvert $(testpath("ijulia.ipynb")) --to markdown`)
+
+# these are the tests to get code coverage
+# nothing like a little piracy when testing in a headless setup
+RCall.ijulia_init()
+R"plot(1:10, 1:10)"
+# throws a method error when running headless !
+@test_throws(MethodError, RCall.ijulia_displayplots())
```
palday, line 35: @ararslan if you have a better idea, I'd be keen. I considered pirating (in the tests) `display(::MIME, Vector{UInt8})` but that seemed even worse

ararslan: I don't think I fully understand the context. What's the desired/typical behavior? If you want to test that it throws a method error then this seems fine...

ararslan (approved):
I know basically nothing about IJulia and have used it no more than twice ever, I believe, so if it's a meaningful review you're after, I'm not your guy. 😛 However, I have this rubber stamp next to me and the ink is still wet...

test/ijulia.jl
palday: When called in a Jupyter notebook, this will result in the R plot (written to a temporary file) being read in and then `display`ed using the appropriate MIME type. But when called in a plain text, headless environment, you get a method error. I'm not too worried about it since we're testing the actual behavior separately (via the `nbconvert` stuff above), but not in a way that codecov detects it (due to the nesting of processes). So the run here is just re-running the same code that's in the notebook so that codecov gets it.

PR #545, opened 2024-09-03
use 'min' option in julia-setup

palday:


ViralBShah:
@palday what should be done to merge? Do you have commit access?

palday:
I needed to figure out why CI started failing and then got distracted by other things. I'll take a look again and see if I can figure it out.

PR #558, opened 2025-03-19
Add versioninfo() helper and a bug report template

frankier:
Since there are different ways to install RCall.jl, this might speed up triage.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,41 @@
+---
+name: Bug report
+about: Report a suspected bug or incompatibility in RCall.jl
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Please try to create as small a reproduction as possible.
+
+ * How did you install RCall.jl and configure the R location?
+ * Did you install any R packages or other packages? Using what method? 
+ * Finally what code did you run?
+
+Please include this information even if it relates to an earlier issue which has been closed.
+
+**Expected and actual behavior**
+Describe what happened versus what you expected to happen.
+
+**Version information**
+Please include the output of follow commands in the REPL:
+
+```julia
+using RCall
+RCall.debuginfo()
+```
+
+In case you have trouble running the above, you can instead include the output of:
+```julia
+using RCall
+reval("sessionInfo()")
+versioninfo()
+```
+
+**Additional context**
+Add any other context about the problem here.
--- a/src/RCall.jl
+++ b/src/RCall.jl
@@ -56,6 +56,7 @@ if Rhome == ""
     __precompile__(false)
 end
 
+include("debuginfo.jl")
 include("types.jl")
 include("Const.jl")
 include("methods.jl")
--- a/src/debuginfo.jl
+++ b/src/debuginfo.jl
@@ -1,0 +1,37 @@
+"""
+Print some useful information about the current RCall.jl setup.
+If `all` is false, only RCall.jl information is printed, otherwise
+information about your Julia and R setup is printed as well.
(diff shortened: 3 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,29 @@
...
+ * Did you install any R packages or other packages? Using what method? 
+ * Finally what code did you run?
+
+Please include this information even if it relates to an earlier issue which has been closed.
+
+**Expected and actual behavior**
+Describe what happened versus what you expected to happen.
+
+**Version information**
+Please include the output of `using RCall; RCall.versioninfo()` if possible. Otherwise please include `Main.versioninfo()` from Julia and `sessionInfo()` from R.
```
palday, line 26:
```suggestion
Please include the output of `using RCall; RCall.versioninfo()` if possible. Otherwise please include `versioninfo()` from Julia and `sessionInfo()` from R.
```

No need to qualify stuff in `Main` when running interactively and if you're not running interactively, then you'll need to load InteractiveUtils ...

palday: Actually, on second thought, I think we should just simplify this a bit:

````suggestion
Please include the output of running the follow commands in the REPL:

```julia
using RCall
reval("sessionInfo()")
versioninfo()
```
````

And then we can cut out the additional `versioninfo` file below (more on that down there)

File: src/versioninfo.jl [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+    println("$(prefix)RCall.jl version $version")
+    println("$(prefix)R location configured with Preferences.jl: $Rhome_set_as_preference")
+    println("$(prefix)RCall.jl managed R using Conda.jl: $conda_provided_r")
+    println("$(prefix)Rhome: $Rhome")
+    println("$(prefix)libR: $libR")
+    if all
+        println()
+        println("Julia information:")
+        buffer = IOBuffer()
+        Main.versioninfo(buffer)
```
palday, line 21: package code should not generally call out to `Main` -- `Main` doesn't exist in some sense from the package's perspective and it can't be listed as a dependency. `versioninfo` is actually defined in `InteractiveUtils`, so if we were doing to do this, I would define a stub

```julia
function versioninfo end
```

in the main package and then define an appropriate package extension that defines the full method when `InteractiveUtils` will be loaded, which is generally the case when the user is running stuff interactively, but not the case if a user is running e.g. 

```sh
julia -e'using RCall; RCall.versioninfo()'
```

palday (commented):
I like the idea of an issue template, but I don't think we need to make any code changes to support it. :smile:

src/versioninfo.jl
frankier: Okay, I can move it to a package extension, but I think it's still useful to keep it, and have one function which shows everything also for people to check the current state themselves. The idea is to show where R has come from and how it's managed by RCall.jl. This could be extended later to try and detect some of these recurring problems that cause people to end up with broken environments.

Perhaps `debuginfo()` could be better.

palday: Yeah, I think `debuginfo` is a nicer name for this use case. :smile: Please put it in a package extension (`RCallInteractiveUtilsExt`) and ping me for re-review.

You'll have to add `InteractiveUtils` to the test environment for testing.

frankier: I tried out what you suggested with the package extension. It seems to only work after explicitly running `using InteractiveUtils` within the REPL which seems undesirable.

I'm not totally sure with what the problem with using `Main` from a module is. Is there some information somewhere telling us that this is not allowed? Does it not work for you? I've found that for developer tools that are intended to be installed into people's global environment, it's fine to grab them from `Main`. I've seen this pattern about the place too. Maybe this is an exception to the rule?

I've updated it with a `isdefined(...)` check which should help to prevent incorrect usage in the common case.

PR #559, opened 2025-03-19
Fix CondaPkg test on Windows

frankier:
This tries to fix the Windows tests with the current CondaPkg. The test is a bit fragile, but hopefully this should be robust enough now.

I also found an unused/incomplete test, which I've deleted.


test/installation/swap_to_prefs_and_condapkg.jl
```diff
@@ -42,8 +42,9 @@ RCall = CondaPkg.withenv() do
     Pkg.build("RCall")
     Base.require(Main, :RCall)
 end
-expected = joinpath("x", ".CondaPkg", "env", "lib", "R")[2:end]
-if occursin(expected, RCall.Rhome)
+expected_condapkg = joinpath("x", ".CondaPkg", "x")[2:end-1]
+expected_env_lib_r = joinpath("x", "env", "lib", "R", "x")[2:end-1]
```
palday, line 46:
```suggestion
expected_condapkg = joinpath("x", ".CondaPkg", "x")[2:(end-1)]
expected_env_lib_r = joinpath("x", "env", "lib", "R", "x")[2:(end-1)]
```

palday (commented):
Thanks, @frankier! We can probably ignore failures on nightly / investigate them in a follow-up, but do you understand what's causing the failure on Julia 1 / R-release on Windows?

frankier:
>do you understand what's causing the failure on Julia 1 / R-release on Windows?

Fixed it now. The reason is an update to CondaPkg to use Pixi on Windows, which changes the path around a little bit. I don't have Windows handy so was using CI to develop my test.

palday (approved):
The failures on nightly aren't blocking for this, but it's still concerning to see a segfault ....

Thanks @frankier for all your hard work on improving support for different platforms!

PR #560, opened 2025-04-12
Remove RException showerror method which supresses backtraces

frankier:
This method definition suppresses the stacktrace for all `RException` exceptions. Can we remove this so that it's easier to determine where an error comes from if there are e.g. lots of `revals` or `R_str`s?

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -54,7 +54,6 @@ end
 
 
 abstract type RException <: Exception end
-showerror(io::IO, e::RException, bt; backtrace=false) = showerror(io ,e)
 
 # status = 2
 struct RParseIncomplete <: RException


palday:
Can you add a concrete example of something that's bad now and which is better after this proposal?

frankier:
Good point. Would be good if I could follow my own issue template(!)

Say we have `backtrace_example.jl` where an R exception will occur in the function `bar`:

```julia
using RCall

function foo()
    R"""
    print("hello")
    """
end

function bar()
    R"""
    stop("error")
    """
end

function main()
    foo()
    bar()
end

main()
```

Before:

```
$ julia --project=. backtrace_example.jl
[1] "hello"
ERROR: LoadError: REvalError: Error: error
Backtrace:
1: stop("error")
in expression starting at /home/frankier/sources/RCall.jl/backtrace_example.jl:20
```

Note that this doesn't tell me which function the exception occurred in. I have to try to figure it out. Note that line 20 is just the call to `main()` so doesn't help.

After:

```
$ julia --project=. backtrace_example.jl
[1] "hello"
ERROR: LoadError: REvalError: Error: error
Backtrace:
1: stop("error")
Stacktrace:
 [1] handle_eval_stderr(; as_warning::Bool)
   @ RCall ~/sources/RCall.jl/src/io.jl:173
 [2] handle_eval_stderr
   @ ~/sources/RCall.jl/src/io.jl:167 [inlined]
 [3] reval_p(expr::Ptr{LangSxp}, env::Ptr{EnvSxp})
   @ RCall ~/sources/RCall.jl/src/eval.jl:103
 [4] reval_p(expr::Ptr{RCall.ExprSxp}, env::Ptr{EnvSxp})
   @ RCall ~/sources/RCall.jl/src/eval.jl:119
 [5] reval(str::String, env::RObject{EnvSxp})
   @ RCall ~/sources/RCall.jl/src/eval.jl:136
 [6] macro expansion
   @ ~/sources/RCall.jl/src/macros.jl:74 [inlined]
 [7] bar
   @ ~/sources/RCall.jl/backtrace_example.jl:10 [inlined]
 [8] main()
   @ Main ~/sources/RCall.jl/backtrace_example.jl:17
 [9] top-level scope
   @ ~/sources/RCall.jl/backtrace_example.jl:20
in expression starting at /home/frankier/sources/RCall.jl/backtrace_example.jl:20
```

Now there's a full stacktrace can read from the bottom to find that the exception is thrown from `bar()`.

palday (approved):

PR #562, opened 2025-04-14
Pin R version to 4.4 for building docs

frankier:
This is just in case the fix takes a while.


palday:
@frankier thanks! I'm onboard with this, but can you restrict this to just the docs in this PR? I think we might want to pin the R version in the doc build anyway. For the main CI, I'm open to a temporary pin, but feel like it should be a separate PR (that we can then revert separately).

frankier:
Okay. Split done.

PR #561, opened 2025-04-14
Fix SET_SYMVALUE removal (broken)

frankier:
This is how far I got trying to fix the removal of `SET_SYMVALUE` in R 4.5. Internally this assignment should do the same thing as the unpublished internal  API `SET_SYMVALUE`, but the test seems to die without a bracktrace so a probably a segfault. I can try to figure it out at some point but need to relearn how. I'm posting my progress here in case there's any obvious advice/fix.

PR #564, opened 2025-04-16
use public API `Rf_defineVar` instead of internal `SET_SYMVALUE`

palday:
R 4.5 introduced a few changes that break RCall:
- stricter internal vs. external API boundaries, which we felt with our call to `SET_SYMVALUE`
- apparently an optimization to the way empty vectors are stored, which broke our handling of them by messing up pointer alignment (at least on x86-64)

Failures on nightly are unrelated: #565

closes #561

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,12 +4,12 @@ concurrency:
   cancel-in-progress: true
 on:
   push:
-    branches: [master]
+    branches: [main]
     tags: ["*"]
   pull_request:
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - R ${{ matrix.R }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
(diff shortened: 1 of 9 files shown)


ararslan (approved):
I can't speak to whether this is correct but it looks reasonable at least.

Could be worth organizing the commits logically and merging without squashing so that the functional change here can be reverted more easily should it ever need to be in the future.

palday:
For some reason, CI on R release/4.5 is unbelievably slow at every stage:
- the `setup-r` step
- the Julia precompile step
- the installation testing step

which means that we don't actually ever complete the job even after nearly two hours. In #568, I'm exploring this issue further. In the meantime, I've tested R 4.5 on a Windows 10 VM and everything worked and I'm setting the R version for Windows CI to 4.4. Maybe this is just an intermittent thing until various mirrors and runners scale up capacity, but I'm going to go ahead and merge this now.

PR #569, opened 2025-04-18
REPL completion fix on 1.12

palday:
Multiple sources of breakage
- https://github.com/JuliaLang/julia/pull/54800; fix adapted from https://github.com/JuliaLang/IJulia.jl/pull/1147
- https://github.com/JuliaLang/julia/pull/57192; some related discussion https://github.com/JuliaLang/julia/issues/58189

Seems like #588 also helped with calling the function pointers from a background thread!

PR #570, opened 2025-04-18
Update RPrompt.jl

stevengj:
Fix REPL completion for https://github.com/JuliaLang/julia/pull/54800, similar to https://github.com/JuliaLang/IJulia.jl/pull/1147.

cc @palday.


stevengj:
Closing in favor of #569.

PR #571, opened 2025-04-22
improve hygiene of rimport macro

palday:
The previous version included an `eval`. 🤔

### DIFF
--- a/src/namespaces.jl
+++ b/src/namespaces.jl
@@ -100,16 +100,20 @@ macro rimport(x, args...)
 end
 
 """
-Load all exported functions/objects of an R package to the current module. Almost equivalent to
+    @rlibrary(package)
+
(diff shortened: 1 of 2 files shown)


src/namespaces.jl
```diff
@@ -100,16 +100,20 @@ macro rimport(x, args...)
...
 ```
 """
 macro rlibrary(x)
-    m = gensym("RCall")
-    quote
-        $(esc(m)) = rimport($(QuoteNode(x)))
-        Core.eval(@__MODULE__, Expr(:using, Expr(:., :., $(QuoteNode(m)))))
-    end
+    tmp = gensym("RCall")
+    ex = Expr(:block,
```
ararslan, line 115:
```suggestion
    ex = Expr(:block, __source__,
```
I think you'll need this in order for backtraces to point to the `@rlibrary` call site.

ararslan (approved):

PR #578, opened 2025-09-19
fix rlibrary-marcro test on Julia 1.12

palday:
Julia 1.12 introduced more strict world age semantics for global bindings, which requires the explicit use of a `eval` when using the `rlibrary` macro within a nested scope and then immediately depending on its results.

> [!NOTE]
> This does not fix the REPL issue on 1.12, so tests will still fail on nightly/RC.


ararslan:
...huh. Linux support is apparently broken?

PR #584, opened 2025-10-05
Repair examples of RCall.RClass

dmbates:
Documentation fix only.  The examples of `RClass` used only `RCall` instead of `RCall.RClass`.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -517,10 +517,10 @@ RObject(x::RObject) = x
 Representation of R Class.
 
 Examples:
-- `RCall{:logical}`
-- `RCall{:integer}`
-- `RCall{:numeric}`
-- `RCall{:character}`
+- `RCall.RClass{:logical}`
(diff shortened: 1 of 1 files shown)


palday (approved):
Failures are unrelated -- they're from something in the latest version of IJulia

JamesWrigley:
Heyo, IJulia maintainer here :) Sorry about the breakage, this is from the big v1.30 refactor. The error is occurring because there's no active kernel, but of course it makes sense to register a callback with IJulia without requiring a kernel to be running. Will fix it in the next release.

JamesWrigley:
Should be fixed in v1.30.6: https://julialang.github.io/IJulia.jl/stable/changelog/#[v1.30.6](https://github.com/JuliaLang/IJulia.jl/releases/tag/v1.30.6)-2025-10-06

palday:
wow @JamesWrigley that was fast, thanks!

